PR #4, opened 2023-09-28
Use cmdline for registration and login

liam-sbhoo:
## Development
- use command line for registration and login
- add a [TabPFNDemo.ipynb](https://github.com/liam-sbhoo/tabpfn-client/blob/use-cmdline-for-registration-and-login/TabPFNDemo.ipynb) (to demonstrate local and remote tabpfn)
- add GitHub workflow for testing

## Note
Perhaps play with [TabPFNDemo.ipynb](https://github.com/liam-sbhoo/tabpfn-client/blob/use-cmdline-for-registration-and-login/TabPFNDemo.ipynb) a little when reviewing :)
- since the bwcloud server is slower than expected (unknown issue), the request to server would timeout sometimes - retrying might help.


tabpfn_client/tabpfn_classifier.py
```diff
@@ -155,45 +155,61 @@ def predict_proba(self, X):
...
+    def indent(text: str) -> str:
+        indent_factor = 2
+        indent_str = " " * indent_factor
+        return textwrap.indent(text, indent_str)
 
     prompt = "\n".join([
         "",
         "Welcome to TabPFN!",
         "",
         "Sadly you are not logged in yet.",
```
noahho, line 169: Please change this text to: "You are not logged in yet."

noahho:
What do you think about moving the access token to ~/.tabpfn/config (instead of ./access_token.txt) - imo this looks more standard and is more descriptive - someone seeing a file just called access_token.txt doesn't know what for?

noahho:
I got a server disconnected error for tabpfn.predict_proba once. Could you add something that retries the request 5 times and only then raises an error?

noahho:
Can you add a logout function?

noahho:
I couldn't comment on this because its been there before, in the client:
`
def predict_proba(self, X):
        check_is_fitted(self)
        X = common_utils.serialize_to_csv_formatted_bytes(X)
        response = self.httpx_client.post(
            url=self.server_endpoints.predict_proba.path,
            headers={"Authorization": f"Bearer {self.access_token}"},
            params={"per_user_train_set_id": self.last_per_user_train_set_id_},
            files=common_utils.to_httpx_post_file_format([
                ("x_file", X)
            ])
        )
`

It seems like we are transmitting the entire file each time - can we add a function that just transmits the dataset key in case its already uploaded? We have all that functionality already right? That would probably make the server communication faster.

liam-sbhoo:
> What do you think about moving the access token to ~/.tabpfn/config (instead of ./access_token.txt) - imo this looks more standard and is more descriptive - someone seeing a file just called access_token.txt doesn't know what for?

Good idea! Will do this.

liam-sbhoo:
> Can you add a logout function?

@noahho Could you elaborate on the user case for this?

liam-sbhoo:
> I couldn't comment on this because its been there before, in the client:
> 
> ```
> def predict_proba(self, X):
>         check_is_fitted(self)
>         X = common_utils.serialize_to_csv_formatted_bytes(X)
>         response = self.httpx_client.post(
>             url=self.server_endpoints.predict_proba.path,
>             headers={"Authorization": f"Bearer {self.access_token}"},
>             params={"per_user_train_set_id": self.last_per_user_train_set_id_},
>             files=common_utils.to_httpx_post_file_format([
>                 ("x_file", X)
>             ])
>         )
> ```
> 
> It seems like we are transmitting the entire file each time - can we add a function that just transmits the dataset key in case its already uploaded? We have all that functionality already right? That would probably make the server communication faster.

I think to achieve this, we would need to implement a few more APIs - we need additional APIs that:
- check if the dataset exists in the database (if yes, return the reference id)
- use stored dataset for prediction and prediction_proba

Do you think we could wait a little for this optimization? I was thinking to get the MVP up and running first.

noahho:
> > Can you add a logout function?
> 
> @noahho Could you elaborate on the user case for this?

This is a good point. There is not really a user case at the moment honestly. It's just such standard functionailty and would only require deleting the access token and resetting internal params? My user case was that I wanted to test creating a new user, but I couldn't since I was already logged in.

liam-sbhoo:
> > > Can you add a logout function?
> > 
> > 
> > @noahho Could you elaborate on the user case for this?
> 
> This is a good point. There is not really a user case at the moment honestly. It's just such standard functionailty and would only require deleting the access token and resetting internal params? My user case was that I wanted to test creating a new user, but I couldn't since I was already logged in.

What if I create a function to just clear all internal params? like tabpfn_classifier.reset() or tabpfn_classifier.clear()

noahho:
> > > > Can you add a logout function?
> > > 
> > > 
> > > @noahho Could you elaborate on the user case for this?
> > 
> > 
> > This is a good point. There is not really a user case at the moment honestly. It's just such standard functionailty and would only require deleting the access token and resetting internal params? My user case was that I wanted to test creating a new user, but I couldn't since I was already logged in.
> 
> What if I create a function to just clear all internal params? like tabpfn_classifier.reset() or tabpfn_classifier.clear()

Yes that sounds good, and clear the file as well?!

liam-sbhoo:
I have resolved most of the issues, while having the following as TODOs (either under the MVP or next phase):

1. Optimize for data uploading (e.g. avoid uploading if the dataset exists in the database, [link to discussion](https://github.com/liam-sbhoo/tabpfn-client/pull/4#issuecomment-1746465423))
2. More robust execution of API (retry calling the APIs if connection error occurs, [link to discussion](https://github.com/liam-sbhoo/tabpfn-client/pull/4#issuecomment-1744968594))

What do you think?

noahho:
That sounds great :)

noahho (approved):

PR #7, opened 2023-10-25
Integrate gdpr api

liam-sbhoo:
# Development
- Add support for GDPR APIs :)
- Refactoring

# Demonstration of GDPR API usage
``` python
from tabpfn_client import tabpfn_classifier, UserDataClient

tabpfn_classifier.init(use_server=True)
data_client = UserDataClient()

# all available API(s) are as followed:
data_client.get_data_summary()  # to get summary of uploaded data
data_client.download_all_datasets(save_dir)  # get a zip file of all uploaded data
data_client.delete_dataset(certain_dataset_uid)  # to delete given train set / test set
data_client.delete_all_datasets()  # to delete all uploaded datasets
data_client.delete_user_account()  # to delete the user account (incl. the datasets)
```
# Note
I couldn't think of a cool name for "UserDataClient". Let me know if you've got anything, I will be glad to change it :)


liam-sbhoo:
Hey Sam and Noah, please also take a look at this client PR when you have the time. Thanks :)

tabpfn_client/server_config.yaml
```diff
@@ -14,30 +20,61 @@ endpoints:
...
+
+  delete_all_datasets:
+    path: "/delete_all_datasets/"
+    methods: [ "DELETE" ]
+    description: "Delete all datasets (both train set and test set)"
+
+  delete_user_account:
+    path: "/delete_user_account/"
+    methods: [ "DELETE" ]
+    description: "Delete user account"
```
noahho, line 79: Add: alongside all data associated with account

noahho (approved):
Hey Liam, thanks for the great PR!
I couldn't think of a better word for UserDataClient either..
maybe @SamuelGabriel want to look over it as well, but its looking good for me.

tabpfn_client/service_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+        except RuntimeError as e:
+            logging.error(f"Failed to delete all datasets: {e}")
+            raise e
+
+        logging.info(f"Deleted datasets: {deleted_datasets}")
+
+        return deleted_datasets
+
+    def delete_user_account(self):
+        confirm_pass = PromptAgent.prompt_confirm_password_for_user_account_deletion()
```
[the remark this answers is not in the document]
liam-sbhoo: Yup it does! :)

tabpfn_client/service_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+        except RuntimeError as e:
+            logging.error(f"Failed to delete user account: {e}")
+            raise e
+
+        PromptAgent.prompt_account_deleted()
+
+
+class InferenceClient(ServiceClientWrapper):
+    """
+    Wrapper of ServiceClient to handle inference, including:
```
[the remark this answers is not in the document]
liam-sbhoo: I wanted to have a clearer distinction (or encapsulation) of the responsibility - so my thought was...
- InferenceClient should handle all the low-level configurations and workflow (especially dealing with ServiceClient()
- RemoteTabPFNClassifier should handle the logic regarding sklearn interface (which I think would be helpful in the coming compatibility task).

tabpfn_client/server_config.yaml
liam-sbhoo: Done :)

tabpfn_client/tabpfn_classifier.py
```diff
@@ -1,69 +1,58 @@
...
 
 
 g_tabpfn_config = TabPFNConfig()
 
 
-def init(
-        use_server=True,
-):
+def init(use_server=True):
     global g_tabpfn_config
```
SamuelGabriel, line 30: given that we now already have something that this interacts with on the level above, it might be nice to move this function to be a member of tabpfn_client directly instead, what do you think?

SamuelGabriel (approved):
Thanks a lot :) This looks pretty gdpr compliant 😆 

I only reviewed it superficially, sorry. Maybe we could talk a bit more in depth about the codebase in our meeting next week.

One general point: I think we should very carefully think about a modularization s.t. we can share a lot of code between local and server TabPFN. But how you did it is fine for now, I think.

tabpfn_client/tabpfn_classifier.py
liam-sbhoo: I'm not sure if I get you correctly here. But we can also briefly discuss this on Monday :)

liam-sbhoo: Will do this in this notion ticket [link](https://www.notion.so/Preliminary-Package-client-5fd3cf39290a45d3a94363cc4d31eefa?pvs=4)

PR #8, opened 2023-11-03
Enable tabpfn classifier config

liam-sbhoo:
# Development
Allow the user to explicitly specify the TabPFN config for the remote TabPFN.

## Workflow
No changes to how the user interacts with the libraries.
```python
from tabpfn import tabpfn_classifier

tabpfn_classifier.init()

# The following now works for local and remote
tabpfn = tabpfn_classifier.TabPFNClassifier(N_ensemble_configurations=100)
```

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -1,13 +1,12 @@
 from pathlib import Path
 import httpx
 import logging
-import copy
+import json
 
 import numpy as np
 from omegaconf import OmegaConf
 
 from tabpfn_client.tabpfn_common_utils import utils as common_utils
-
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +25,7 @@ class ServiceClient:
         self.server_config = SERVER_CONFIG
         self.server_endpoints = SERVER_CONFIG["endpoints"]
         self.base_url = f"{self.server_config.protocol}://{self.server_config.host}:{self.server_config.port}"
-        self.httpx_timeout_s = 30   # temporary workaround for slow computation on server side
+        self.httpx_timeout_s = 30  # temporary workaround for slow computation on server side
         self.httpx_client = httpx.Client(
             base_url=self.base_url,
             timeout=self.httpx_timeout_s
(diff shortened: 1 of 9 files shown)


liam-sbhoo:
Rebase before merging
- [ ] rebase

SamuelGabriel:
Could you please ping me again when you have merged the other PR, s.t the diff is cleaner :)

liam-sbhoo:
@SamuelGabriel @noahho 
This PR is updated and ready for review. Thanks :)

tabpfn_client/remote_tabpfn_classifier.py
```diff
@@ -35,26 +35,46 @@ def __init__(
...
+        # self.average_logits = average_logits
+        # self.categorical_features = categorical_features
+        # self.optimize_metric = optimize_metric
         self.seed = seed
-        self.transformer_predict_kwargs_init = transformer_predict_kwargs_init
+        # self.transformer_predict_kwargs_init = transformer_predict_kwargs_init
         self.multiclass_decoder = multiclass_decoder
 
         self.inference_handler = inference_handler
+        self.inference_handler.config_tabpfn(
```
SamuelGabriel, line 58: It would be good to allow users to change the settings later on, just like they can with he classical TabPFN. That means we should only when `predict_proba` is called create the config.

You could hand this config over to the inference_handler in predict_proba as well.

tabpfn_client/client.py
```diff
@@ -88,7 +87,7 @@ def upload_train_set(self, X, y) -> str:
         train_set_uid = response.json()["train_set_uid"]
         return train_set_uid
 
-    def predict(self, train_set_uid: str, x_test):
+    def predict(self, train_set_uid: str, x_test, tabpfn_config: dict = None):
```
SamuelGabriel, line 90: I think it would be better to change the API s.t. you only have a predict_proba endpoint, as this really is the superset of predict. Then we could save a lot of complication especially in the layers of this PR.

SamuelGabriel (changes requested):
I think it would be good to change two things about this PR:
- only use a predict_proba endpoint in the API, and transform for predict locally
- allow changes to the settings later on

tabpfn_client/client.py
liam-sbhoo: Agreed.

I kept `predict` instead of `predict_proba` though, would that be okay? Just a slight preference of mine (we could also change this) Rationale being: calling `predict_proba(with_proba=false)` seems strange (like ordering cheese burger but without cheese) :D

liam-sbhoo:
Reimplemented in #21

PR #16, opened 2024-02-19
Announce new features and version compatibility check

davidotte:
Update client for retrieving new messages of a user and supporting the client version compatibility check of the server.

More detail:
- Add current client version to the header of each request
- Update error raising such that e.g. exceptions from the server regarding the client version are shown the user (which was not the case before)
- Retrieve and print messages during init() call

I rebased to the fix_tests_client branch from Anshul, which is not merged yet.


quick_test.py
```diff
@@ -32,7 +32,7 @@
 
     else:
         tabpfn_classifier.init()
-        tabpfn = TabPFNClassifier()
+        tabpfn = TabPFNClassifier(model="latest_tabpfn_hosted")
         # print("checking estimator", check_estimator(tabpfn))
         print(X_train.shape[0]*100)
         tabpfn.fit(np.repeat(X_train, 100, axis=0), np.repeat(y_train, 100, axis=0))
```
SamuelGabriel, line 38: Since there was a bug here that we did not spot from the last PR, could you please add a small test that this file runs to the tests

tabpfn_client/client.py
```diff
@@ -15,6 +14,15 @@
 SERVER_CONFIG = OmegaConf.load(SERVER_CONFIG_FILE)
 
 
+def get_client_version() -> str:
+    try:
+        return version('tabpfn_client')
+    except PackageNotFoundError:
+        # Package not found, should only happen during development. Simply return a version number that works
+        # for development.
+        return '5.5.5'
```
SamuelGabriel, line 23: How can I fix this to get the real version number? I can imagine that in the future we might care about trying our code with the right version number.

Is it fixed with a `pip install -e .`? If so please add this to the comment

tabpfn_client/client.py
```diff
@@ -117,19 +126,25 @@ def predict(self, train_set_uid: str, x_test):
 
         return np.array(response.json()["y_pred"])
 
-    def error_raising(self, response, method_name):
+    def error_raising(self, response, method_name, only_version_check=False):
```
SamuelGabriel, line 129: Can't we simply do all checks always? Why not?

tabpfn_client/tabpfn_classifier.py
```diff
@@ -130,34 +132,42 @@ def fit(self, X, y):
...
+                "feature_shift_decoder": self.feature_shift_decoder,
+                # "normalize_with_test": self.normalize_with_test,
+                # "average_logits": self.average_logits,
+                # "categorical_features": self.categorical_features,
+                # "optimize_metric": self.optimize_metric,
+                "seed": self.seed,
+                # "transformer_predict_kwargs_init": self.transformer_predict_kwargs_init,
+                "multiclass_decoder": self.multiclass_decoder
+            }
+            #classifier_cfg = {}
```
SamuelGabriel, line 155: these diffs are from anshul's pr, right?

SamuelGabriel (changes requested):
Thank you very much! **Very** tiny things. I did not look too thoroughly at the tests, but think it is ready to be merged.

tabpfn_client/client.py
davidotte: All checks always would mean that always a RuntimeError is raised when something is wrong. It might be desirable that, e.g. when the client tries to retrieve messages from the server but there is some server error, the user still can proceed without getting the error.

SamuelGabriel: Do you think so? Are there cases like this? I think this is kinda dangerous. I rather get some arbitrary error, then my predict call just not returning anything without any reason, no? Maybe I am understanding something false, though.

davidotte: Yes, but in the example of a predict call it would raise the error because we do not set only_version_check to True. In the other cases where only_version_check is True, e.g. during registration, the program will still terminate at the moment but with a more customized exception.
I think the question here is whether it is better to raise an error anyway and if necessary chatch it in the service_wrapper function to handle it, or to ignore the error in this function on purpose because we know we will handle it later. What do you think?

davidotte: @liam-sbhoo what is your opinion on this?

liam-sbhoo: I would suggest always checking the version actually. In general, it's simpler.
And it could also make debugging easier too: in your last example, let's say the `prediction` api doesn't work. Having the check tells us that we are working on compatible version when debugging.

liam-sbhoo, line 129: If this is a private method, you could prefix it with `_` as a practice, i.e. `self._error_raising`.
Also, `error_raising` sounds vague to me. Could you give it a better name?

tabpfn_client/tabpfn_classifier.py
```diff
@@ -52,6 +52,8 @@ def init():
             # prompt for login / register
             PromptAgent.prompt_and_set_token(user_auth_handler)
 
+        PromptAgent.prompt_retrieved_messages(user_auth_handler.retrieve_messages())
```
liam-sbhoo, line 55: I suppose if there's no new messages, it wouldn't print anything right?
Maybe include a comment to clarify this.

tabpfn_client/client.py
```diff
@@ -117,19 +126,25 @@ def predict(self, train_set_uid: str, x_test):
...
         if response.status_code != 200:
             load = None
             try:
                 load = response.json()
             except Exception:
                 pass
+            if (response.status_code == 403 and load and load.get("detail").startswith("Client version too old")
+                    or response.status_code == 400 and load and load.get("detail").startswith("Client version")):
+                raise RuntimeError(load.get("detail"))
+            if only_version_check:
```
liam-sbhoo, line 139: We could use some comments in the following lines to better demonstrates the logic.

liam-sbhoo (changes requested):
Comment:
- Including comments are strategic spots might help readers to understand your thought process.
- There may be room for improvement (for better robustness)

(Others) TODO:
[] Rebase on main - easier to review

tabpfn_client/client.py
davidotte: I just used the the function definition as it was before and expanded it for the version check. But yes, you're right, I will try to improve this. :)
Regarding the version check, it is checked always, right? With the new only_version_check argument we can just decide if we also want to check for other errors.

davidotte: We already check the version always, right? With the new only_version_check argument we can just decide if we also want to check for other errors.

tabpfn_client/client.py
```diff
@@ -117,19 +126,25 @@ def predict(self, train_set_uid: str, x_test):
...
 
-    def error_raising(self, response, method_name):
+    def error_raising(self, response, method_name, only_version_check=False):
         if response.status_code != 200:
             load = None
             try:
                 load = response.json()
             except Exception:
                 pass
+            if (response.status_code == 403 and load and load.get("detail").startswith("Client version too old")
```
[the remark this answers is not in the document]
davidotte: Hm, good point. I just found that there is a status code "426: Upgrade required", I think this might be a good choice

quick_test.py
davidotte: I get an error when I try to run this file:
RuntimeError: Fail to call upload_train_set with error: {'detail': 'The training dataset provided is too large, as the number of examples/lines is 38099 which is greater than the maximum number of lines (9999).'}
Is this intended?

tabpfn_client/client.py
davidotte: Yes it works, I will add it to the comment.

davidotte:
Hey, I just rebased to main and applied all the requested changes :)

tabpfn_client/client.py
```diff
@@ -251,6 +278,7 @@ def login(self, email: str, password: str) -> str | None:
             data=common_utils.to_oauth_request_form(email, password)
         )
 
+        self._validate_response(response, "login", only_version_check=True)
```
SamuelGabriel, line 281: please set only_version check to false here

tabpfn_client/client.py
SamuelGabriel: i think we kinda talked passed each other: as a general principle let us try to always communicate as much as we know about an error to the user. in the places you use `only_version_check` it is fine, only please set it to false on login (i added a comment there too). let's resolve this.

quick_test.py
SamuelGabriel: i might have added this during debugging. please change this s.t. it runs through, i.e. decrease number of examples to something super small like 100

SamuelGabriel:
This all looks nice!

this is good to be merged (without another look) after you:
1. set `only_version_check=False` in line 281 client.py
2. make sure the `quick_test.py` runs through without error, sorry for adding this to this PR, but I think it is just nice if this works.
3. use "426: Upgrade required

Thanks :)

quick_test.py
SamuelGabriel: ah I sea, then the test is unnecessary!

davidotte:
@SamuelGabriel Done.

tabpfn_client/client.py
```diff
@@ -113,23 +122,37 @@ def predict(self, train_set_uid: str, x_test):
...
+    def _validate_response(response, method_name, only_version_check=False):
+        # If status code is 200, no errors occurred on the server side.
+        if response.status_code == 200:
+            return
+
+        # Read response.
+        load = None
+        try:
+            load = response.json()
+        except Exception:
```
liam-sbhoo, line 139: Sorry I forgot to mention this, could we only catch known `Exception` and log the non-expected one? It seems a little dangerous.

I got the following from ChatGPT:
```python
try:
    response_payload = response.json()
except json.JSONDecodeError as e:
    logging.error(f"Failed to parse JSON from response in {method_name}: {e}")
```

liam-sbhoo: Remember to log it.

tabpfn_client/tests/unit/test_client.py
```diff
@@ -27,6 +28,14 @@ def test_try_connection_with_invalid_server(self, mock_server):
         mock_server.router.get(mock_server.endpoints.root.path).respond(404)
         self.assertFalse(self.client.try_connection())
 
+    @with_mock_server()
+    def test_try_connection_with_outdated_client(self, mock_server):
```
liam-sbhoo, line 32: In general, a better practice for the naming the test cases should include:
- test condition
- expected result

In this case, you could call it `def test_try_connection_with_outdated_client_raises_xxx`

ps. I figured that the old test cases were following this naming convention, but let's start applying this now :)

liam-sbhoo (approved):
Only very minor things, good to merge after this improvement.

SamuelGabriel (approved):
LGTM

PR #17, opened 2024-03-18
Update user registration

davidotte:
This PR introduces the following things:
- Simplify registration/login process by allowing retries
- Ask users some additional questions during registration.

## Allowing retries
Update registration and login functions such that they return if the login/registration was successfull and the response of the server. If the attempt was not successfull, print the response of the server (e.g. 'Weak Password') and restart the registration/login process.

## Asking additional questions
After the registration process that we had before is finished, ask the user a few more questions, which are optional. We can fine-tune the exact wording. The information entered is then sent to the server in a second request. The user is already added to the database after the first request (which we already had before). I thought this was simplest way to do it, if we send everything in one request, all the answers to the questions will have to be entered again if there was a problem at the beginning, e.g. with the password.

If you have further ideas/thoughts on this please let me know :)

### DIFF
--- a/tabpfn_client/prompt_agent.py
+++ b/tabpfn_client/prompt_agent.py
@@ -57,7 +57,7 @@ class PromptAgent:
                 if is_created:
                     break
                 print(cls.indent("User registration failed:" + error_message) + "\n")
-
+            cls.prompt_add_user_information(user_auth_handler)
             print(cls.indent("Account created successfully!") + "\n")
 
         elif choice == "2":
@@ -98,6 +98,29 @@ class PromptAgent:
             raise RuntimeError("Invalid choice")
 
         return choice.lower() == "y"
+
+    @classmethod
+    def prompt_add_user_information(cls, user_auth_handler: "UserAuthenticationClient"):
+        print(cls.indent("To help us tailor our support and services to your needs, we have a few optional questions. "
+                         "Feel free to skip any question by leaving it blank.") + "\n")
+        company = input(cls.indent("Where do you work?: "))
+        role = input(cls.indent("What is your role?: "))
+        use_case = input(cls.indent("What do you want to use TabPFN for?: "))
+
+        contact_via_email = input(cls.indent("Can we reach out to you via email to support you? (y/n):"))
+        is_valid_choice = False
+        for _ in range(3):
+            if contact_via_email.lower() not in ["y", "n"]:
+                contact_via_email = input(cls.indent("Invalid choice, please enter 'y' or 'n': "))
+            else:
+                is_valid_choice = True
+                break
+        if not is_valid_choice or contact_via_email.lower() == "n":
+            contact_via_email = False
+        else:
+            contact_via_email = True
+
+        user_auth_handler.add_user_information(company, role, use_case, contact_via_email)
 
     @classmethod
     def prompt_reusing_existing_token(cls):
--- a/tabpfn_client/prompt_agent.py
+++ b/tabpfn_client/prompt_agent.py
@@ -136,16 +136,14 @@ class PromptAgent:
         choice = input(cls.indent(prompt))
 
         # retry for 3 attempts until valid choice is made
-        is_valid_choice = False
-        for _ in range(max_attempts):
+        for _ in range(max_attempts - 1):
             if choice.lower() not in choices:
                 choices_str = ", ".join(f"'{item}'" for item in choices[:-1]) + f" or '{choices[-1]}'"
                 choice = input(cls.indent(f"Invalid choice, please enter {choices_str}: "))
             else:
-                is_valid_choice = True
                 break
 
-        if not is_valid_choice:
+        if not choice.lower() in choices:
             raise RuntimeError("Invalid choice")
 
         return choice.lower()
--- a/tabpfn_client/prompt_agent.py
+++ b/tabpfn_client/prompt_agent.py
@@ -37,31 +37,39 @@ class PromptAgent:
             #validation_link = input(cls.indent("Please enter your secret code: "))
             validation_link = "tabpfn-2023"
             # create account
-            email = input(cls.indent("Please enter your email: "))
+            while True:
+                email = input(cls.indent("Please enter your email: "))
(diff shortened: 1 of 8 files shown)


tabpfn_client/prompt_agent.py
```diff
@@ -23,45 +23,55 @@ def prompt_welcome(cls):
...
-                "",
-                "Please enter your password: ",
-            ])
-
-            password = getpass.getpass(cls.indent(password_req_prompt))
-            password_confirm = getpass.getpass(cls.indent("Please confirm your password: "))
-
-            user_auth_handler.set_token_by_registration(email, password, password_confirm, validation_link)
-
+            while True:
```
SamuelGabriel, line 41: I think if we work on this already, it would be even better (and a considerable convenience boost) if we have this loop around the mail and the password prompt separately s.t. if I mistype my mail I get a warning about this straight away. For this: run the email validator here, too:

```python
email_info = email_validator.validate_email(
                email, test_environment=test_mode
            )
```

and also have a loop to check the passwords are equal.

tabpfn_client/prompt_agent.py
```diff
@@ -23,45 +23,55 @@ def prompt_welcome(cls):
...
 
+        # Login
         elif choice == "2":
             # login to account
-            email = input(cls.indent("Please enter your email: "))
-            password = getpass.getpass(cls.indent("Please enter your password: "))
-
-            user_auth_handler.set_token_by_login(email, password)
-
+            while True:
```
SamuelGabriel, line 67: here it is fine to just have one loop

tabpfn_client/prompt_agent.py
```diff
@@ -74,22 +84,23 @@ def prompt_terms_and_cond(cls) -> bool:
             "By using TabPFN, you agree to the following terms and conditions:",
             "Do you agree to the above terms and conditions? (y/n): ",
         ])
+        choice = cls._choice_with_retries(t_and_c, ["y", "n"])
+        return choice.lower() == "y"
```
SamuelGabriel, line 88: I like the choice with retries a lot, you do not need the `lower()` here anymore though

tabpfn_client/prompt_agent.py
```diff
@@ -74,22 +84,23 @@ def prompt_terms_and_cond(cls) -> bool:
...
-            if choice.lower() not in ["y", "n"]:
-                choice = input(cls.indent("Invalid choice, please enter 'y' or 'n': "))
-            else:
-                is_valid_choice = True
-                break
+    @classmethod
+    def prompt_add_user_information(cls, user_auth_handler: "UserAuthenticationClient"):
+        print(cls.indent("To help us tailor our support and services to your needs, we have a few optional questions. "
+                         "Feel free to skip any question by leaving it blank.") + "\n")
+        company = input(cls.indent("Where do you work?: "))
```
SamuelGabriel, line 94: leave the colons (:) away, as you already have a question mark, in all three lines

tabpfn_client/prompt_agent.py
```diff
@@ -115,3 +126,24 @@ def prompt_confirm_password_for_user_account_deletion(cls) -> str:
     @classmethod
     def prompt_account_deleted(cls):
         print(cls.indent("Your account has been deleted."))
+
+    @classmethod
+    def _choice_with_retries(cls, prompt: str, choices: list, max_attempts: int = 3) -> str:
+        """
+        Prompt text and give user predefined number of attempts to select one of the possible choices. If valid choice
+        is selected, return choice in lowercase, otherwise raise RuntimeError.
+        """
```
SamuelGabriel, line 135: add assert to only have lower cased choices
```suggestion
        """
        assert all(c.lower() == c for c in choices), "Choices need to be lower case."
```

tabpfn_client/prompt_agent.py
```diff
@@ -115,3 +126,24 @@ def prompt_confirm_password_for_user_account_deletion(cls) -> str:
     @classmethod
     def prompt_account_deleted(cls):
         print(cls.indent("Your account has been deleted."))
+
+    @classmethod
+    def _choice_with_retries(cls, prompt: str, choices: list, max_attempts: int = 3) -> str:
```
SamuelGabriel, line 131: why have a max attempts? I think this could also go for ever, no?

SamuelGabriel (changes requested):
Very cool functionality added and I do like the `_choice_with_retries`, the main thing that I would think should be improved is to have a separate loop for the email address validity, the password following our rules and the passwords matching

tabpfn_client/prompt_agent.py
davidotte: Thanks for the feedback, I have now implemented this and it is definitely a much better user experience. :)
One thing to add: We can also have the case that the entered email already exists. This might not happen very often, but I added an extra request that is sent to the server to also check this directly when the email is entered.

tabpfn_client/prompt_agent.py
davidotte: Yeah, you're right, I changed it. But not sure if it will ever make a difference :grinning:

SamuelGabriel: nah me neither, but it is one less trajectory in the code to think about. Simpler :)

tabpfn_client/prompt_agent.py
SamuelGabriel: yeah that is great, I wondered why you implemented email validation on the server side. good thinking!

SamuelGabriel (approved):
Can I ask for one more tiny thing: now that we have None as an option in the DB, could you please set empty inputs to None. Right now we would still save empty strings in the DB, if I read the code right.

Otherwise: This looks great! Thanks a lot!

I used approve as I don't think there is another review necessary after fixing the None thing above.

davidotte:
Perfect, thanks. :)
This is already done implicitly because only the strings that are not empty are sent to the server. In the test db it works with the other values being null.

SamuelGabriel:
ah great! this is ready to be merged without a second review

PR #20, opened 2024-04-25
Refactor Additional User Info on Client

anshulg954:
### Change Description

Similar points as described on Server

**Details** (add details if your pull request is more complicated and harder to understand from the code alone)
API hit prevented.

na

@SamuelGabriel  

---

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -263,7 +263,8 @@ class ServiceClient:
             email: str,
             password: str,
             password_confirm: str,
-            validation_link: str
+            validation_link: str,
+            additional_info: dict
     ) -> tuple[bool, str]:
         """
         Register a new user with the provided credentials.
@@ -274,6 +275,7 @@ class ServiceClient:
         password : str
         password_confirm : str
         validation_link: str
+        additional_info : dict
 
         Returns
         -------
@@ -285,8 +287,16 @@ class ServiceClient:
 
         response = self.httpx_client.post(
             self.server_endpoints.register.path,
-            params={"email": email, "password": password, "password_confirm": password_confirm,
-                    "validation_link": validation_link}
+            params={
+                "email": email, 
+                "password": password, 
+                "password_confirm": password_confirm,
+                "validation_link": validation_link, 
+                "company": additional_info["company"], 
+                "role": additional_info["role"], 
+                "use_case": additional_info["use_case"], 
+                "contact_via_email": additional_info["contact_via_email"]
+            }
         )
 
         self._validate_response(response, "register", only_version_check=True)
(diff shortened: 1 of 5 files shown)


tabpfn_client/client.py
```diff
@@ -275,8 +277,16 @@ def register(
...
         response = self.httpx_client.post(
             self.server_endpoints.register.path,
-            params={"email": email, "password": password, "password_confirm": password_confirm,
-                    "validation_link": validation_link}
+            params={
+                "email": email, 
+                "password": password, 
+                "password_confirm": password_confirm,
+                "validation_link": validation_link, 
+                "company": additional_info["company"],
```
SamuelGabriel, line 285: i think the better thing to use here is `**additional_info` if you do not have an argument against that

tabpfn_client/prompt_agent.py
```diff
@@ -51,7 +51,7 @@ def prompt_and_set_token(cls, user_auth_handler: "UserAuthenticationClient"):
         # Registration
         if choice == "1":
             # validation_link = input(cls.indent("Please enter your secret code: "))
-            validation_link = "tabpfn-2023"
+            validation_link = "tabpfn-test"
```
SamuelGabriel, line 54: why is this different now? this won't work in production then, right?

tabpfn_client/prompt_agent.py
```diff
@@ -126,7 +126,12 @@ def prompt_add_user_information(cls, user_auth_handler: "UserAuthenticationClien
         )
         contact_via_email = True if choice_contact == "y" else False
 
-        user_auth_handler.add_user_information(company, role, use_case, contact_via_email)
```
SamuelGabriel, line 129: you need to delete UserAuthenticationClient.add_user_information function still, right?

tabpfn_client/service_wrapper.py
```diff
@@ -39,13 +39,11 @@ def set_token_by_registration(
             email: str,
             password: str,
             password_confirm: str,
-            validation_link: str
+            validation_link: str,
+            additional_info: dict
     ) -> tuple[bool, str]:
 
-        is_created, message = self.service_client.register(email, password, password_confirm, validation_link)
-        if is_created:
```
SamuelGabriel, line 46: we still do want to set the token here, don't we?

SamuelGabriel (changes requested):
thanks for the PR. i think there are a few things to be done:
i) the cleanup of the old API is not completely done, I think, e.g. in service_wrapper.py
ii) you changed some settings, that I do not think you want to change

tabpfn_client/client.py
anshulg954: Changed to this **additional_info and it is working fine as expected.

tabpfn_client/prompt_agent.py
anshulg954: No changes in this

anshulg954:
Ready for merge

SamuelGabriel (approved):

PR #21, opened 2024-05-04
Pass hyper params from client

SamuelGabriel:
### Change Description

In this PR, I start passing all arguments to the server. To make this simpler I removed the local tabpfn functionality for now but kept the main code s.t. we can easily re-enable it later on. This was simpler, because the local tabpfn has very different parameters from the server one.

DEPENDENCY: this PR depends on https://github.com/automl/tabpfn-server/pull/48/ to filter out the "model" kwarg

yes, i removed local

no

@liam-sbhoo @anshulg954 @davidotte 

---


liam-sbhoo (commented):
Looking great in overall!

Just some comments/questions:
- regarding the model args, why not pop the `model` args in the client instead and keep the server untouched? Then we'll have one less change to make when we have the feature enabled.
- can we add some test cases, concerning the presence of `tabpfn_config`? Just a very minimal safe guard to ensure that we take `tabpfn_config`

SamuelGabriel:
Tests: yeah, good point. I will add a test ensuring the defaults are send.

"model" arg: at some point we will need this model arg anyways, as we will train new models and want users to be able to choose, so I thought why not now. I should make it backward compatible on the backend, though.

tabpfn_client/prompt_agent.py
```diff
@@ -51,7 +51,7 @@ def prompt_and_set_token(cls, user_auth_handler: "UserAuthenticationClient"):
         # Registration
         if choice == "1":
             # validation_link = input(cls.indent("Please enter your secret code: "))
-            validation_link = "tabpfn-2023"
+            validation_link = "tabpfn-test"
```
liam-sbhoo, line 54: The production is still using `tabpfn-2023`, right?

liam-sbhoo: In the server, should we just keep the `validation_link` the same for all `DEV`, `PREPROD`, and `PROD`?
Or actually, do we still need this `validation_link` feature?

liam-sbhoo (commented):
Some issues on the `validation_link`.

tabpfn_client/prompt_agent.py
SamuelGabriel: let's keep it synced. i will update it to be 2023 in DEV too

PR #23, opened 2024-05-19
Add ability to reset password

davidotte:
This PR belongs to this [server PR](https://github.com/automl/tabpfn-server/pull/51).

When a user tries to login and the login fails, he gets the opportunity to either retry to login or to request a password reset email.

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -355,6 +355,22 @@ class ServiceClient:
 
         return response.json()["requirements"]
 
+    def send_reset_password_email(self, email: str) -> tuple[bool, str]:
+        """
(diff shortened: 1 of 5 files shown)


SamuelGabriel (changes requested):
This looks generally very good. I just tested it locally and added a few changes that I think would help here: https://github.com/automl/tabpfn-client/pull/24

I added the message you wanted to have reminding the user of the email, removed the tabpfn dependency.
And most importantly: I update the (min) version number (also on the backend in my PR there) as the interface between client and backend changed. Could you add these changes (merge the branch), then this should be good to go :)

SamuelGabriel (approved):
LGTM

PR #25, opened 2024-06-01
Add Ruff

davidotte:
### Change Description

Add ruff and pre-commit to client repo (analogous to server repo)

**Details** (add details if your pull request is more complicated and harder to understand from the code alone)

yes

@liam-sbhoo 

---

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,6 +5,16 @@ on:
       - main
 
 jobs:
+  check_python_linting:
+    name: Ruff Linting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: chartboost/ruff-action@v1
+        with:
+          src: "./"
+          version: 0.3.3
+
   test:
     name: Run unit and integration tests
     runs-on: ubuntu-latest
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -9,11 +9,12 @@ jobs:
     name: Ruff Linting
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: chartboost/ruff-action@v1
         with:
           src: "./"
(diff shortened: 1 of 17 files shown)


.github/workflows/pull_request.yml
```diff
@@ -5,6 +5,16 @@ on:
       - main
 
 jobs:
+  check_python_linting:
+    name: Ruff Linting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
liam-sbhoo, line 12: Maybe use v4? That's the latest version available and `tabpfn-server` is using v4 too.

.github/workflows/pull_request.yml
```diff
@@ -5,6 +5,16 @@ on:
...
+      - uses: chartboost/ruff-action@v1
+        with:
```
liam-sbhoo, line 14: Could you confirm if the current GitHub workflow check for formatting issues? I have a vague memory that our github test only checks for linting issues.

This used to be a bug of ruff-action package, and it seems they finally fixed it here (https://github.com/ChartBoost/ruff-action?tab=readme-ov-file#use-ruff-format).

README.md
```diff
@@ -43,3 +43,10 @@ tabpfn.fit(X_train, y_train)
 tabpfn.predict(X_test)
 # or you can also use tabpfn.predict_proba(X_test)
 ```
+
+# Development
```
liam-sbhoo, line 47: Can you also mention this (https://docs.astral.sh/ruff/integrations/) to encourage developers to have the extensions installed in the editor?

liam-sbhoo (changes requested):
Thank you for picking this up! Just some minor comments.

.github/workflows/pull_request.yml
davidotte: Okay yeah, it looks like we have to add this argument such that formatting is checked. But it seems like as soon I add the argument, linting is not checked any longer, at least the output logs suggest that. So I just added it with and without the argument :thinking:

PR #26, opened 2024-06-04
Regression

SamuelGabriel:
### Change Description

I added regression in correspondence to Liam's PR https://github.com/automl/tabpfn-server/pull/49, which should be merged before this PR.

No new dep's.

@liam-sbhoo @davidotte 

## Breaking changes

Yes, based on PR https://github.com/automl/tabpfn-server/pull/49

Yes, I break the user facing API slightly, removing `tabpfn_client.tabpfn_classifier.init/reset` because I think these are more global. They are still available in `tabpfn_client` directly.

---


tabpfn_client/estimator.py
SamuelGabriel, line 1: this is shown as a new file, but the code in TabPFNClassifier (only added doc string), PreprocessorConfig was not touched.

So you would only need to review the TabPFNRegressor bit.

tabpfn_client/config.py
SamuelGabriel, line 1: this was simply moved here from tabpfn_classifier, no new logic added

tabpfn_client/estimator.py
```diff
@@ -0,0 +1,351 @@
...
+        self.add_fingerprint_features = add_fingerprint_features
+        self.cancel_nan_borders = cancel_nan_borders
+        self.super_bar_dist_averaging = super_bar_dist_averaging
+        self.subsample_samples = subsample_samples
+
+    def fit(self, X, y):
+        # assert init() is called
+        if not config.g_tabpfn_config.is_initialized:
+            raise RuntimeError(
+                "tabpfn_client.init() must be called before using TabPFNClassifier"
```
liam-sbhoo, line 318: Should be `TabPFNRegressor` instead.

tabpfn_client/estimator.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+
+        if config.g_tabpfn_config.use_server:
+            try:
+                assert (
+                    self.model == "latest_tabpfn_hosted"
+                ), "Only 'latest_tabpfn_hosted' model is supported at the moment for tabpfn_classifier.init(use_server=True)"
```
liam-sbhoo, line 325: Should be `TabPFNRegressor` instead.

tabpfn_client/estimator.py
```diff
@@ -0,0 +1,351 @@
...
+            except AssertionError as e:
+                print(e)
+            config.g_tabpfn_config.inference_handler.fit(X, y)
+            self.fitted_ = True
+        else:
+            raise NotImplementedError(
+                "Only server mode is supported at the moment for tabpfn_classifier.init(use_server=False)"
```
liam-sbhoo, line 332: Should be `TabPFNRegressor` instead.

tabpfn_client/client.py
```diff
@@ -135,7 +142,15 @@ def predict(self, train_set_uid: str, x_test, tabpfn_config: dict | None = None)
 
         self._validate_response(response, "predict")
 
-        return np.array(response.json()["y_pred_proba"])
+        result = response.json()[task]
+
```
liam-sbhoo, line 146: Could you include comment to tell why we handle the result based on the types? i.e. mention we could get classification or regression results. Might be helpful for other developers :)

liam-sbhoo: OR

Would it be better if we move this logic to `TabPFNClassifier` and `TabPFNRegressor`?

tabpfn_client/server_config.yaml
liam-sbhoo, line 1: I think these changes should not be added.

liam-sbhoo (commented):
Thanks for the PR!
Just one thing about the parsing of results (and some minor comments).

p.s. thanks for the comments regarding your changes (e.g. moving code around, renaming and such), it was a great help!

tabpfn_client/server_config.yaml
SamuelGabriel: sorry! that was a mistake!

tabpfn_client/estimator.py
SamuelGabriel: done

tabpfn_client/estimator.py
SamuelGabriel: done

SamuelGabriel:
I changed all your requested things. Is it ready to be merged (ideally we should merge this and the server's partner PR before the meeting today)

liam-sbhoo:
LTGM, thanks!

liam-sbhoo (approved):

tabpfn_client/client.py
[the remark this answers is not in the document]
SamuelGabriel: good catch, delted it

PR #30, opened 2024-07-04
Add data checks in client.py

Sathya98:
### Change Description

Added a function check_training_data that checks the X and y tables for basic integrity and length inconsistencies before sending them over to the tabpfn server

Also added a unit test in test_client.py that tests the function with the breast cancer dataset

## Breaking changes

---

### DIFF
--- a/tabpfn_client/estimator.py
+++ b/tabpfn_client/estimator.py
@@ -6,7 +6,6 @@ import numpy as np
 from tabpfn_client import init
 from sklearn.base import BaseEstimator, ClassifierMixin, RegressorMixin
 from sklearn.utils.validation import check_is_fitted
-from sklearn.utils import check_consistent_length
 
 from tabpfn_client import config
 
@@ -422,7 +421,8 @@ def validate_data_size(X: np.ndarray, y: np.ndarray | None = None):
 
     # check if the number of samples is consistent (ValueError)
     if y is not None:
-        check_consistent_length(X, y)
+        if X.shape[0] != y.shape[0]:
+            raise ValueError("X and y must have the same number of samples")
 
     # length and feature assertions
     if X.shape[0] > MAX_ROWS:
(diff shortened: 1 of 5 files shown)


Sathya98:
Yes

liam-sbhoo:
Please also setup your environment for ruff formatting according to `Development` section of the README. With this, you'll probably pass the failed checks.

liam-sbhoo:
Good to go after the slight refactor in test and correcting the format. Thanks!!

liam-sbhoo:
I did some modification here:
- moving the data size check to `estimator`, instead of `ServiceClient` (responsibility-wise is clearer)
- add more test cases

tabpfn_client/estimator.py
```diff
@@ -393,3 +409,23 @@ def _model_name_to_path(self, model_name: str) -> str:
...
+    Check the integrity of the training data.
+    - check if the number of rows between X and y is consistent
+        if y is not None (ValueError)
+    - check if the number of rows is less than MAX_ROWS (ValueError)
+    - check if the number of columns is less than MAX_COLS (ValueError)
+    """
+
+    # check if the number of samples is consistent (ValueError)
+    if y is not None:
+        check_consistent_length(X, y)
```
SamuelGabriel, line 425: since the rest of the code relies on `.shape` to exist, can you simply use that here too s.t. it is directly understandable what you are doing

SamuelGabriel (changes requested):
Just one tiny little change I would request than it is good to go!

tabpfn_client/estimator.py
liam-sbhoo: Alright, done!

SamuelGabriel (approved):
LGTM

tabpfn_client/estimator.py
SamuelGabriel: thanks

PR #31, opened 2024-07-15
fix class remappings

SamuelGabriel:
### Change Description

We had a bug that, the following code yielded wrong results, because we forgot to remap classes that are not contiguous, but e.g. [0,4,8]:

```
import numpy as np
m = tabpfn_client.TabPFNClassifier()
X = np.random.rand(100,2)
y_cls = (X.mean(1) > .5) * 2
m.fit(X,y_cls)
X_test = np.random.rand(100,2)
pred_test = m.predict(X_test)
```
`pred_test` with the current client is from the set {0,1}, but it should be {0,2}.

## Breaking changes

---

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -170,11 +170,11 @@ class ServiceClient:
         try:
             load = response.json()
         except json.JSONDecodeError as e:
-            logging.error(f"Failed to parse JSON from response in {method_name}: {e}")
+            logging.info(f"Failed to parse JSON from response in {method_name}: {e}")
 
         # Check if the server requires a newer client version.
         if response.status_code == 426:
-            logger.error(
+            logger.info(
                 f"Fail to call {method_name}, response status: {response.status_code}"
             )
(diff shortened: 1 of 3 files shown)


tabpfn_client/client.py
```diff
@@ -170,11 +172,11 @@ def _validate_response(response, method_name, only_version_check=False):
         try:
             load = response.json()
         except json.JSONDecodeError as e:
-            logging.error(f"Failed to parse JSON from response in {method_name}: {e}")
+            logging.info(f"Failed to parse JSON from response in {method_name}: {e}")
```
liam-sbhoo, line 175: Why changed it to `info` though?

liam-sbhoo (approved):
Just a question about the logging level, thanks!

tabpfn_client/client.py
SamuelGabriel: As otherwise our outputs look ugly with these messages in them, even though it is not really an error, since it is intended behavior that our ValueErrors are caught that way.

PR #33, opened 2024-08-15
Store last train set uid in classifier/regressor class

davidotte:
### Change Description

Currently we have the following incorrect behavior:
model1.fit(X1,Y1)
model2.fit(X2,Y2)
model1.predict will now use X2, Y2 instead of X1, Y1

This PR fixes this problem by storing the last_train_set_uid in the Classifier/Regressor class instead of the InferenceClient class. This allows each classifier/regressor instance to keep track of the corresponding last train set it was fitted on.

-
@SamuelGabriel 

## Breaking changes

No
No
---


SamuelGabriel (approved):
Nice change, well done!

PR #36, opened 2024-09-14
Remove outdated assertion on model support

liam-sbhoo:
### Change Description

I think this assertion is outdated.

**Addition / Fix:**
- `TabPFNRegressor.list_available_models()`
- add model support
- fix unnecessary welcoming message everytime we call fit.

**I have tested it (checked from Vertex log, make sure the model specified in client is being used).**

NO

## Breaking changes

---

### DIFF
--- a/tabpfn_client/estimator.py
+++ b/tabpfn_client/estimator.py
@@ -232,9 +232,18 @@ class TabPFNClassifier(BaseEstimator, ClassifierMixin):
 
 
 class TabPFNRegressor(BaseEstimator, RegressorMixin):
+
+    _AVAILABLE_MODELS = [
+        "default",
+        "2noar4o2",
+        "5wof9ojf",
+        "09gpqh39",
+        "wyl4o83o",
+    ]
+
     def __init__(
         self,
-        model: str = "latest_tabpfn_hosted",
+        model_path: str = "default",
         n_estimators: int = 8,
         preprocess_transforms: Tuple[PreprocessorConfig, ...] = (
             PreprocessorConfig(
@@ -310,7 +319,7 @@ class TabPFNRegressor(BaseEstimator, RegressorMixin):
                 If in 0 to 1, the value is viewed as a fraction of the training set size.
         """
 
-        self.model = model
+        self.model_path = self._model_name_to_path(model_path)
         self.n_estimators = n_estimators
         self.preprocess_transforms = preprocess_transforms
         self.feature_shift_decoder = feature_shift_decoder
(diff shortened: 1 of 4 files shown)


SamuelGabriel (approved):
LGTM

tabpfn_client/tests/unit/test_tabpfn_classifier.py
```diff
@@ -58,10 +58,6 @@ def test_init_remote_classifier(
             mock_server.endpoints.retrieve_greeting_messages.path
         ).respond(200, json={"messages": []})
 
-        mock_server.router.get(mock_server.endpoints.protected_root.path).respond(
```
liam-sbhoo, line 61: Ideally, we don't need this here.
This line used to pass because we are calling `init()` multiple times, and the subsequent `init()` will call this API.

liam-sbhoo: @SamuelGabriel

tabpfn_client/tests/unit/test_tabpfn_regressor.py
```diff
@@ -56,10 +56,6 @@ def test_init_remote_regressor(
             mock_server.endpoints.retrieve_greeting_messages.path
         ).respond(200, json={"messages": []})
 
-        mock_server.router.get(mock_server.endpoints.protected_root.path).respond(
```
liam-sbhoo, line 59: Same as https://github.com/automl/tabpfn-client/pull/36/files#r1760022912

tabpfn_client/tests/unit/test_tabpfn_classifier.py
SamuelGabriel: what do you mean? we don't need the "protected" call? I think so too as almost any action will do a protected call. removing it will save latency. do you mean that?

liam-sbhoo: This API is a dummy endpoint that requires authentication. One use case is to check the user can login successfully or not. All the other API doesn't go through.

tabpfn_client/estimator.py
```diff
@@ -232,9 +232,17 @@ def predict_proba(self, X):
...
+        "2noar4o2",
+        "5wof9ojf",
+        "09gpqh39",
+        "wyl4o83o",
+    ]
+
     def __init__(
         self,
-        model: str = "latest_tabpfn_hosted",
+        model_path: str = "default",
```
SamuelGabriel, line 245: can you please stick to the "model" name

liam-sbhoo:
I renamed `model_path` to `model` (although some pre-processing required, TabPFN takes both `model` and `model_path` and they meant different things).
Added some test case too.

PR #41, opened 2024-09-26
Add two minor fixes for data download

davidotte:
### Change Description

Two minor fixes:
- Client version header was missing in one request
- Add extra if statement in data validation for data download

@SamuelGabriel 

## Breaking changes

no

no

no

---

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -193,6 +193,8 @@ class ServiceClient:
         # Read response.
         load = None
         try:
+            # This if clause is necessary for streaming responses (e.g. download) to
+            # prevent httpx.ResponseNotRead error.
             if not response.is_closed:
                 response.read()
             load = response.json()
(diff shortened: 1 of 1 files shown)


tabpfn_client/client.py
```diff
@@ -193,6 +193,8 @@ def _validate_response(
         # Read response.
         load = None
         try:
+            if not response.is_closed:
```
SamuelGabriel, line 196: what does this do? could you add a comment why you do this?

SamuelGabriel (approved):
looks good generally

tabpfn_client/client.py
davidotte: This is to make sure that the response is read before response.json() gets called. When dealing with a streaming response (like during data download), I sometimes got an httpx.ResponseNotRead error without this if block. I will add a comment.

SamuelGabriel: Oh interesting. Thank you!

PR #42, opened 2024-10-02
Increase timeout, fix model argument

liam-sbhoo:
#### Patch
- increase httpx timeout
- fix bug:  `TabPFNClassifier`'s `model` argument didn't work.

@SamuelGabriel 

## Breaking changes

---

### DIFF
--- a/tabpfn_client/estimator.py
+++ b/tabpfn_client/estimator.py
@@ -134,7 +134,7 @@ class TabPFNModelSelection:
             )
 
     @classmethod
-    def model_name_to_path(
+    def _model_name_to_path(
         cls, task: Literal["classification", "regression"], model_name: str
     ) -> str:
         cls._validate_model_name(model_name)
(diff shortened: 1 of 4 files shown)


tabpfn_client/estimator.py
```diff
@@ -212,8 +212,8 @@ def fit(self, X, y):
         if config.g_tabpfn_config.use_server:
             try:
                 assert (
-                    self.model == "latest_tabpfn_hosted"
```
SamuelGabriel, line 215: Could you please also remove this assert and add the other models, like you did for regression?

SamuelGabriel (approved):
Could you add the extra models for cls as well pls. That would be nice, even without this is ready to be merged.

tabpfn_client/estimator.py
liam-sbhoo: Okay done :)

SamuelGabriel (approved):
Looks good but for one name :) Thanks for this PR! These are important features :)

tabpfn_client/estimator.py
```diff
@@ -114,7 +114,46 @@ def to_dict(self):
...
+    @classmethod
+    def _validate_model_name(cls, model_name: str) -> None:
+        if model_name != "default" and model_name not in cls._AVAILABLE_MODELS:
+            raise ValueError(
+                f"Invalid model name: {model_name}. "
+                f"Available models are: {cls.list_available_models()}"
+            )
+
+    @classmethod
+    def model_name_to_path(
```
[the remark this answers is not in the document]
liam-sbhoo: Yeah totally, this should stay private.

PR #43, opened 2024-10-17
Hashing based Verification Token Handling 

anshulg954:
### Change Description

It is an enhancement functionality where we ask user to input the verification token instead of verifying it with a link.

@noahho 

## Breaking changes

It did not break anything but a few users complained that our emails landed in spam. It is an update where we try to minimize the landing of emails in spam. Corresponding server PR: https://github.com/automl/tabpfn-server/pull/76

No, it enhances the user interface.
---

### DIFF
--- a/tabpfn_client/config.py
+++ b/tabpfn_client/config.py
@@ -33,15 +33,16 @@ def init(use_server=True):
                 "TabPFN is inaccessible at the moment, please try again later."
             )
 
-        is_valid_token_set = UserAuthenticationClient.try_reuse_existing_token()
+        is_valid_token, access_token = (
+            UserAuthenticationClient.try_reuse_existing_token()
+        )
 
-        if isinstance(is_valid_token_set, bool) and is_valid_token_set:
+        if is_valid_token:
             PromptAgent.prompt_reusing_existing_token()
-        elif (
-            isinstance(is_valid_token_set, tuple) and is_valid_token_set[1] is not None
-        ):
+        elif access_token is not None:
+            # token holds invalid due to user email verification
             print("Your email is not verified. Please verify your email to continue...")
-            PromptAgent.reverify_email(is_valid_token_set[1])
+            PromptAgent.reverify_email(access_token)
         else:
             PromptAgent.prompt_welcome()
             if not PromptAgent.prompt_terms_and_cond():
--- a/tabpfn_client/prompt_agent.py
+++ b/tabpfn_client/prompt_agent.py
@@ -103,18 +103,22 @@ class PromptAgent:
                         )
                     )
             additional_info = cls.prompt_add_user_information()
-            is_created, message = UserAuthenticationClient.set_token_by_registration(
-                email, password, password_confirm, validation_link, additional_info
+            is_created, message, access_token = (
+                UserAuthenticationClient.set_token_by_registration(
+                    email, password, password_confirm, validation_link, additional_info
+                )
             )
             if not is_created:
                 raise RuntimeError("User registration failed: " + str(message) + "\n")
 
             print(
                 cls.indent(
-                    "Account created successfully! To start using TabPFN please click on the link in the verification email we sent you."
+                    "Account created successfully! To start using TabPFN please enter the verification code we sent you by mail."
                 )
                 + "\n"
             )
+            # verify token from email
+            cls._verify_user_email(access_token=access_token)
 
         # Login
         elif choice == "2":
@@ -123,12 +127,17 @@ class PromptAgent:
                 email = input(cls.indent("Please enter your email: "))
                 password = getpass.getpass(cls.indent("Please enter your password: "))
 
-                successful, message = UserAuthenticationClient.set_token_by_login(
-                    email, password
+                access_token, message, status_code = (
+                    UserAuthenticationClient.set_token_by_login(email, password)
                 )
-                if successful:
+                if status_code == 200 and access_token is not None:
                     break
-                print(cls.indent("Login failed: " + message) + "\n")
+                print(cls.indent("Login failed: " + str(message)) + "\n")
+                if status_code == 403:
+                    # 403 implies that the email is not verified
+                    cls._verify_user_email(access_token=access_token)
+                    UserAuthenticationClient.set_token_by_login(email, password)
+                    break
 
                 prompt = "\n".join(
                     [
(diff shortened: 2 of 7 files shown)


anshulg954:
This also fixes the bug in reverify email. @liam-sbhoo

tabpfn_client/config.py
```diff
@@ -44,6 +44,7 @@ def init(use_server=True):
         ):
             print("Your email is not verified. Please verify your email to continue...")
             PromptAgent.reverify_email(is_valid_token_set[1], user_auth_handler)
+            user_auth_handler.set_token(is_valid_token_set[1])
```
tabpfn, line 47: Can you adapt line 38 and replace the array / tuple by two separate variables

A, B = user_auth_handler.try_reuse_existing_token() # with A, B replace by what that means

tabpfn_client/prompt_agent.py
```diff
@@ -100,18 +100,35 @@ def prompt_and_set_token(cls, user_auth_handler: "UserAuthenticationClient"):
...
+                    email, password, password_confirm, validation_link, additional_info
+                )
             )
             if not is_created:
                 raise RuntimeError("User registration failed: " + str(message) + "\n")
 
             print(
                 cls.indent(
-                    "Account created successfully! To start using TabPFN please click on the link in the verification email we sent you."
+                    "Account created successfully! To start using TabPFN please enter the secret key in the verification email we sent you."
```
tabpfn, line 113: To start using TabPFN please enter the verification code we sent to you by mail

tabpfn_client/prompt_agent.py
```diff
@@ -100,18 +100,35 @@ def prompt_and_set_token(cls, user_auth_handler: "UserAuthenticationClient"):
...
+                token = input(cls.indent("Verification Token: "))
+                verified, message = user_auth_handler.verify_email(token)
+                if not verified:
+                    print("\n" + cls.indent(str(message) + "Please try again!") + "\n")
+
+            print(
+                cls.indent(
+                    "Thank you for verifying your email successfully! Your access token is: "
+                )
+                + access_token
```
tabpfn, line 129: Don't show the access token here. Only write:

cls.indent(
                    "Thank you for verifying your email successfully!

tabpfn_client/prompt_agent.py
```diff
@@ -120,12 +137,29 @@ def prompt_and_set_token(cls, user_auth_handler: "UserAuthenticationClient"):
...
                     email, password
                 )
                 if successful:
                     break
-                print(cls.indent("Login failed: " + message) + "\n")
+                print(cls.indent("Login failed: " + str(message)) + "\n")
+                if status_code == 403:
+                    # Verify email
+                    verified = False
+                    while not verified:
```
tabpfn, line 149: Somehow verification is duplicated? Why here and above?

tabpfn_client/prompt_agent.py
tabpfn: To access the token there should be a function get_token and set_token

tabpfn (changes requested):
See the inline comments

tabpfn_client/config.py
anshulg954: Sure. change accomodated, I did this initially to have minimal changes in the inner function try_reuse_existing_token.

tabpfn_client/prompt_agent.py
anshulg954: Done

tabpfn_client/prompt_agent.py
anshulg954: Sure.

anshulg954: Access token thing is in another PR. should we do all in one instead or keep it separate. usually separate notion boards have separate PRs

tabpfn_client/prompt_agent.py
```diff
@@ -120,12 +137,29 @@ def prompt_and_set_token(cls, user_auth_handler: "UserAuthenticationClient"):
...
 
-                successful, message = user_auth_handler.set_token_by_login(
+                successful, message, status_code = user_auth_handler.set_token_by_login(
                     email, password
                 )
                 if successful:
                     break
-                print(cls.indent("Login failed: " + message) + "\n")
+                print(cls.indent("Login failed: " + str(message)) + "\n")
+                if status_code == 403:
```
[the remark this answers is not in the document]
anshulg954: Yes. comment added as well

tabpfn_client/prompt_agent.py
anshulg954: For registration, Login, and reverification. for login we need to set_token using email and password.
However, i have added a separate function now for verification functionality to avoid this - how about this?

anshulg954:
Hi,
I have pushed the changes and added comments wherever necessary. Please also check the corresponding PR on server side.

Thanks

noahho (approved):

SamuelGabriel (changes requested):
I have a minor change proposal. The only question I have left: what do we do if a user is not verified and forgot their password? Both in the case they have the access token for their account and if they do not.

tabpfn_client/service_wrapper.py
```diff
@@ -49,21 +49,22 @@ def set_token_by_registration(
...
 
-    def set_token_by_login(self, email: str, password: str) -> tuple[bool, str]:
-        access_token, message = self.service_client.login(email, password)
+    def set_token_by_login(self, email: str, password: str) -> tuple[bool, str, int]:
+        access_token, message, status_code = self.service_client.login(email, password)
 
         if access_token is None:
-            return False, message
+            return False, message, status_code
+        elif status_code == 200:
```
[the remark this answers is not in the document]
anshulg954: Because if it is 200 we need not prompt the user for the verification token as the user is already verified. However, in the case of 403 we need to reverify email in which we send the user a new verification token, which he can enter in the input prompt and then we set the token there.

anshulg954:
> for

Hi Sam,
In both the cases user can reset the password and then re-login where reverification will take place if the user is not verified.

tabpfn_client/service_wrapper.py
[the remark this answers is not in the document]
SamuelGabriel: yeah i see that you want to do it that way, but I don't understand why

SamuelGabriel:
ok, i see. this sounds good :)

tabpfn_client/service_wrapper.py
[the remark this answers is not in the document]
SamuelGabriel: you can merge this even with this open, but i would like to understand whether this is something that we should fix later on or whether there is an actual reason to handle the access_token differently here

PR #45, opened 2024-10-28
Setter of Access token

anshulg954:
### Change Description

Enhancement to set the access token if the user already has it. 
- The user can now set the token via  init(access_token="access_token"). If the access token is valid we will use it, otherwise the pre-existing prompting will continue.
- Also the user can view his/her token by get_token() directly.
- Additional Information regarding access token can also be found in all the above functionalities.
- Test cases have been updated and added.
- Readme.md has been updated as per functionality.
- Fixed Issues in resetting token. Earlier it was just deleting the file but not clearing the existing cache and was not updating the existing config.
- Also gave a temporary fix for reverify email() which has been fixed thoroughly in hashing PR.  

@liam-sbhoo 
## Breaking changes

No

No
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,14 @@ pip install tabpfn-client
 Import and login
 ```python
 from tabpfn_client import init, TabPFNClassifier
-init()
+
+- To create a new account, simply call the init() or complete the registration process during tabpfn.fit(). 
+
+- Once registered, you will get the information about where your access token is stored, which you can also get by get_token() if it exists.
+
+- Have a valid access token? Directly login using:
+init(access_token="valid_access_token")
+
 ```
 
 Now you can use our model just like any other sklearn estimator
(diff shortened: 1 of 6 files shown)


liam-sbhoo:
Thanks for the work!

I think we could add one more feature, where the user can set the token directly without having to go through the prompting. One example is, `tabpfn.init(access_token='blablabla')`.

liam-sbhoo (changes requested):
Thanks for the work!

I think we could add one more feature, where the user can set the token directly without having to go through the prompting. One example is, tabpfn.init(access_token='blablabla').

noahho:
This is a misunderstaning in how the access token is sued. We only want to set the access token using a functional interface, so by calling tabpfn.set_token(...) not using the prompt interface. The token is useful if you want to programatically login, so I think we don't need any new code, but only add to the README doc how to login using the token

liam-sbhoo (dismissed):
Can you add some test cases? Also perhaps a demonstration on this work.

P.S: When you create a PR, would be nice to provide the reviewer an idea/demonstration of the new workflow.

anshulg954:
Hi Liam!
I did not think it as of a major change the first time I created this PR as it was a few lines of change. However, later forgot to modify the description about the changes. Sincere apologies for the same. 

As per your comments, I have updated the code and added test cases as well keeping the changes to minimal.

noahho (approved):

SamuelGabriel:
I think this PR does not have much new stuff compared to the current main. I would close this, ok? Sorry about this

PR #46, opened 2024-10-29
Add dataset cache

davidotte:
### Change Description

This PR adds a dataset cache to prevent repeated upload of datasets. There are also few changes in the corresponding [server PR](https://github.com/automl/tabpfn-server/pull/80). More concretely, this PR adds the following things:
- Cache system: The DatasetCacheManager keeps track of the cache in form of an OrderedDict, which contains a map of dataset hash to dataset uid for the last 50 uploaded datasets. To avoid specific edge cases (like change of user account or same test set for different train sets), the dataset is hashed together with the user's token and train_set_uid (in the case of the test set). For hashing I decided to use CityHash128 because in my tests it was way faster for large datasets than other common hashing algorithms. The cache is stored in a file in the .tabpfn folder.
- Caching in train set upload: When fit is called and the train set is in the cache, then no request is sent to the server at all.
- Caching in predict: If the test set is cached, then test_set_uid is sent instead of the actual test set. This part involves now a retry mechanism (implemented with the for loop) in case of any failures with cashing. This could for instance be needed if any datasets on server side were deleted by us or the cache file was manually changed. But this should almost never happen. Then, either cashed train_set_uid or test_set_uid can be incorrect, so both datasets are uploaded again.

Please check this PR carefully for the case that I missed any further edge cases :)

yes

@SamuelGabriel 

## Breaking changes

Yes, because now just the cached dataset uids might be sent to server.

---

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -44,8 +44,6 @@ class DatasetUIDCacheManager:
     def __init__(self):
         self.file_path = CACHE_DIR / "dataset_cache"
         self.cache = self.load_cache()
-        self._last_train_X = None
-        self._last_train_y = None
         self.cache_limit = 50
 
     def load_cache(self):
@@ -110,13 +108,6 @@ class DatasetUIDCacheManager:
             self.save_cache()
             return hash_to_delete
         return None
-
-    def temporary_save_train_set(self, X, y):
-        self._last_train_X = X
-        self._last_train_y = y
-
-    def get_temporary_saved_train_set(self):
-        return self._last_train_X, self._last_train_y
 
 
 # Apply the custom filter to the httpx logger
@@ -196,7 +187,6 @@ class ServiceClient:
 
         """
         # Save until prediction for retrying train set upload for the case that anything went wrong with cache.
-        self.dataset_uid_cache_manager.temporary_save_train_set(X, y)
         X_serialized = common_utils.serialize_to_csv_formatted_bytes(X)
         y_serialized = common_utils.serialize_to_csv_formatted_bytes(y)
 
@@ -231,6 +221,8 @@ class ServiceClient:
         x_test,
         task: Literal["classification", "regression"],
         tabpfn_config: dict | None = None,
+        X_train=None,
+        y_train=None,
     ) -> dict[str, np.ndarray]:
         """
         Predict the class labels for the provided data (test set).
@@ -282,9 +274,10 @@ class ServiceClient:
                 ):
                     # Retry by re-uploading the train set
                     self.dataset_uid_cache_manager.delete_uid(train_set_uid)
-                    X_train, y_train = (
-                        self.dataset_uid_cache_manager.get_temporary_saved_train_set()
-                    )
+                    if X_train is None or y_train is None:
+                        raise RuntimeError(
+                            "Train set data is required to re-upload but was not provided."
+                        )
                     train_set_uid = self.upload_train_set(X_train, y_train)
                     params["train_set_uid"] = train_set_uid
                     cached_test_set_uid = None
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -57,17 +57,17 @@ class DatasetCacheManager:
                 return OrderedDict(data)
         return OrderedDict()
 
-    def get_dataset_uid(self, hash):
+    def get_dataset_uid(self, hash: str):
         """
         Retrieves the dataset UID corresponding to the given hash, returns None if dataset not in cache.
         """
-        if hash in self.cache:
+        if str(hash) in self.cache:
             self.cache.move_to_end(hash)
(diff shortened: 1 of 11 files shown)


tabpfn_client/client.py
```diff
@@ -107,20 +186,33 @@ def upload_train_set(self, X, y) -> str:
...
-
-        X = common_utils.serialize_to_csv_formatted_bytes(X)
-        y = common_utils.serialize_to_csv_formatted_bytes(y)
+        # Save until prediction for retrying train set upload for the case that anything went wrong with cache.
+        self._cache_manager.temporary_save_train_set(X, y)
+        X_serialized = common_utils.serialize_to_csv_formatted_bytes(X)
+        y_serialized = common_utils.serialize_to_csv_formatted_bytes(y)
+
+        # Get hash for dataset. Include access token for the case that one user uses different accounts.
+        dataset_hash = str(
```
SamuelGabriel, line 195: I think the hashing could very well go inside the `DatasetUIDCache`, the function should just accept the X,y and extra strings to incorporate (e.g. acccess_token or train_uid)

tabpfn_client/client.py
```diff
@@ -146,29 +238,66 @@ def predict(
             The predicted class labels.
         """
 
-        x_test = common_utils.serialize_to_csv_formatted_bytes(x_test)
+        x_test_serialized = common_utils.serialize_to_csv_formatted_bytes(x_test)
 
-        params = {"train_set_uid": train_set_uid, "task": task}
+        # Get hash for dataset. Include train_set_uid for the case that the same test set was previously used
+        # with different train set. Include access token for the case that a user uses different accounts.
+        dataset_hash = str(
```
SamuelGabriel, line 245: this hashing could also go to the cache manager

tabpfn_client/client.py
```diff
@@ -71,6 +149,7 @@ def __init__(self):
         )
 
         self._access_token = None
+        self._cache_manager = DatasetCacheManager()
```
SamuelGabriel, line 152:
```suggestion
        self._dataset_uid_cache_manager = DatasetCacheManager()
```

tabpfn_client/client.py
```diff
@@ -146,29 +238,66 @@ def predict(
...
+                    train_set_uid = self.upload_train_set(X_train, y_train)
+                    self._cache_manager.add_dataset(old_train_set_hash, train_set_uid)
+                    params["train_set_uid"] = train_set_uid
+                    cached_test_set_uid = None
+                else:
+                    raise  # Re-raise the exception if it's not recoverable
 
-        self._validate_response(response, "predict")
+        else:
+            raise RuntimeError("Failed to get prediction after retries.")
```
SamuelGabriel, line 291:
```suggestion
            raise RuntimeError(f"Failed to get prediction after {max_attempts} retries.")
```

SamuelGabriel (changes requested):
This looks generally nice. There are two weak points that would need to be addressed though, I believe. There seems to be a slight bug, when using multiple estimator objects, and the naming is oftentimes hard to get, as it confuses "Dataset" and "DatasetUID".

tabpfn_client/client.py
```diff
@@ -32,6 +36,80 @@ def filter(self, record):
         return True
 
 
+class DatasetCacheManager:
```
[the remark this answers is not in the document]
davidotte: Thanks for the review, good catch with the bug! I now moved the last train set part to the estimators, I feel like this is more straightforward with our current implementation than including it in the InferenceClient. I also did some renaming to avoid confusion. :)

tabpfn_client/client.py
```diff
@@ -32,6 +36,80 @@ def filter(self, record):
...
+        os.makedirs(os.path.dirname(self.file_path), exist_ok=True)
+        with open(self.file_path, "w") as file:
+            json.dump(self.cache, file)
+
+    def delete_dataset_by_uid(self, dataset_uid: str) -> Optional[str]:
+        """
+        Deletes an entry from the cache based on the dataset UID.
+        """
+        hash_to_delete = None
+        for hash, uid in self.cache.items():
```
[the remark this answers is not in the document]
davidotte: I think we implicitly have this rule. When adding a dataset_uid to the dictionary, the old value is automatically overwritten.

tabpfn_client/client.py
davidotte: I also thought about this, I was a bit hesitant to do that because we hash very different kind of stuff in fit and predict. But I agree with you. I moved the hashing to the DatesetUIDCacheManager and simply combine all arguments for the hash. Still not 100% sure how intuitive my new implementation is

SamuelGabriel: 👍

SamuelGabriel (approved):
Thanks for all the fixes. This PR looks very good now.

PR #49, opened 2024-12-06
test

LeoGrin:
### Change Description

## Breaking changes

---

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-httpx>=0.27.2
+httpx==0.27.2
 omegaconf>=2.3.0
 pandas>=1.3.0
 password-strength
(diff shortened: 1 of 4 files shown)


LeoGrin:
This is a toy pr (no change) to check that the CI breaking is not due to some changes in another PR. Would some of you have an idea of what's causing the error? @SamuelGabriel @davidotte maybe?  Maybe something changed in the CI or something? I get no error when running the tests locally.

davidotte:
No idea, I would also need to investigate :thinking:

noahho:
Could it be the mock server tabpfn-server-wjedmz7r5a-ez.a.run.app has changed? The restx library seems to be stable for a while

noahho:
Locally things are running, but not on the server

LeoGrin:
Okay it seems to be due to version differences in the `httpx` package. Downgrading to `httpx-0.27.2` makes it work locally while the latest version (`0.28.1`) also fails locally. `0.28.1` came out on December 6th, and we don't specify the version in the `requirement.txt` file so the CI uses the latest version.

LeoGrin:
#51 should fix this

PR #54, opened 2024-12-13
Make prediction async

davidotte:
### Change Description

This PR involves the client side for making predictions async. Please also check the corresponding [Server PR](https://github.com/automl/tabpfn-server/pull/92). On the client side, we only need updates in client.py. We now send streaming requests and basically just loop over all messages we receive. Currently, if a duration estimation is received, I print the estimated duration and then show a progress bar that runs for this estimated duration. The question is now if we want to keep this initial print of estimated duration and if we might also want to add a message that is printed if the prediction is taking longer than expected.

Yes. Should I also add version numbers?

@noahho 

## Breaking changes

Yes, PR mentioned above

---

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -317,7 +317,9 @@ class ServiceClient:
                             elif data["error_class"] == "ValueError":
                                 raise ValueError(data["detail"])
                             else:
-                                raise RuntimeError(data["detail"])
+                                raise RuntimeError(
+                                    data["error_class"] + ": " + data["detail"]
+                                )
                 break
             except RuntimeError as e:
                 error_message = str(e)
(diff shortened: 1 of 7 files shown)


tabpfn_client/client.py
```diff
@@ -258,14 +271,52 @@ def predict(
...
+                                progress_bar.n = progress_bar.total
+                                progress_bar.refresh()
+                                progress_bar.close()
+                        elif data["event"] == "error":
+                            if data["error_class"] == "GCPOverloaded":
+                                raise GCPOverloaded(data["detail"])
+                            elif data["error_class"] == "ValueError":
+                                raise ValueError(data["detail"])
+                            else:
+                                raise RuntimeError(data["detail"])
```
noahho, line 318: Add "error_class"

noahho (approved):
Looks good to me

tabpfn_client/client.py
davidotte: What do you mean, add it in the RuntimeError message?

noahho: raise RuntimeError(data["error_class"] + data["detail"])

PR #60, opened 2025-01-04
Store additional user info

davidotte:
### Change Description

This PR adds/changes the following things:

- Ask user for first and last name during registration and send information to the server
- Move to terms and conditions into registration process
- User has to agree to PII, send user agreements to the server

no new dep

@noahho 

## Breaking changes

Yes, see corresponding PR

---

### DIFF
--- a/tabpfn_client/prompt_agent.py
+++ b/tabpfn_client/prompt_agent.py
@@ -198,7 +198,7 @@ class PromptAgent:
     def prompt_terms_and_cond(cls) -> bool:
         t_and_c = "\n".join(
             [
-                "Please refer to our terms and conditions at: https://www.priorlabs.ai/terms-eu-en "
+                "\nPlease refer to our terms and conditions at: https://www.priorlabs.ai/terms-eu-en "
                 "By using TabPFN, you agree to the following terms and conditions:",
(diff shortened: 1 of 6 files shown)


tabpfn_client/prompt_agent.py
```diff
@@ -180,23 +198,53 @@ def prompt_and_set_token(cls):
     def prompt_terms_and_cond(cls) -> bool:
         t_and_c = "\n".join(
             [
-                "Please refer to our terms and conditions at: https://www.priorlabs.ai/terms-eu-en "
+                "\nPlease refer to our terms and conditions at: https://www.priorlabs.ai/terms-eu-en "
```
noahho, line 201: - Can you change this to just https://www.priorlabs.ai/terms please?

noahho (approved):
Great! Did you check if this will be backwards compatible with old accunts?

davidotte:
For old accounts the first and last name fields in the user table will just be empty. And we of course do not have any user agreements stored for them.

tabpfn_client/prompt_agent.py
davidotte: Done

PR #62, opened 2025-01-06
changes to be compatible with new interfaces

LeoGrin:
### Change Description

## Breaking changes

---


quick_test.py
```diff
@@ -20,6 +20,9 @@
     X_train, X_test, y_train, y_test = train_test_split(
         X, y, test_size=0.33, random_state=42
     )
+    # make y text
```
noahho, line 23: Should have noticed before: This filename + documentation isn't great, can you make naming clearer?

tabpfn_client/client.py
```diff
@@ -241,6 +241,7 @@ def predict(
         train_set_uid: str,
         x_test,
         task: Literal["classification", "regression"],
+        predict_params: dict | None = None,
```
noahho, line 244: Much better!

tabpfn_client/estimator.py
```diff
@@ -157,100 +56,91 @@ class TabPFNClassifier(BaseEstimator, ClassifierMixin, TabPFNModelSelection):
...
+            is outside of this range, the model may not perform well. You may ignore
+            our limits to use the model on data outside the pre-training range.
+        inference_precision: "autocast" or "auto", default="auto"
+            The precision to use for inference. This can dramatically affect the
+            speed and reproducibility of the inference.
+        random_state: int or RandomState or RandomGenerator or None, default=None
+            Controls the randomness of the model. Pass an int for reproducible results.
+        inference_config: dict or None, default=None
+            Additional advanced arguments for model interface.
+        paper_version: bool, default=False
```
noahho, line 115: What is this param? Described in the paper?

noahho: Thus, the old tabpfn code

tabpfn_client/estimator.py
```diff
@@ -16,110 +15,10 @@
 MAX_COLS = 500
 
 
-@dataclass(eq=True, frozen=True)
```
noahho, line 19: Are the config options now not documented anywhere? Maybe still good to copy the new ones for reference?

tabpfn_client/estimator.py
LeoGrin: yeah I need to improve the doc for this actually thanks

PR #76, opened 2025-01-14
Clean text in client to fix serialization

LeoGrin:
### Change Description

#70 was not enough to fix the serialization issues, I found some other issues which are solved by replacing `\s+` by a space, and added some more tests.

## Breaking changes

---

### DIFF
--- a/tabpfn_client/estimator.py
+++ b/tabpfn_client/estimator.py
@@ -388,6 +388,11 @@ def _check_paper_version(paper_version, X):
 
 
 def _clean_text_features(X):
+    """
+    Clean text features in the input data. This is used to avoid
+    serialization errors, which happens when the input data contains
+    commas or weird spaces, and to limit the length of the text features.
+    """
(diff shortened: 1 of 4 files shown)


tabpfn_client/estimator.py
```diff
@@ -405,7 +405,13 @@ def _clean_text_features(X):
             pd.to_numeric(X_[col])
         except Exception:
```
noahho, line 406: More specific exception to catch?

tabpfn_client/estimator.py
```diff
@@ -405,7 +405,13 @@ def _clean_text_features(X):
...
             if X_[col].dtype == object:  # only process string/object columns
-                X_[col] = X_[col].str.replace(",", "").str.slice(0, 2500)
+                X_[col] = (
```
noahho, line 408: Add a comment on why we do this?

noahho (approved):
Approved with small suggestions

PR #77, opened 2025-01-16
add validation for nans in y and number of classes > 10

LeoGrin:
### Change Description

Add validation for nans in y, and for number of classes > 10 for classification. This is already checked in the TabPFN local package, but right now be don't pass the error back to the client so the user doesn't see it + it's quite wasteful to send a request when this would fail anyway. In the future it would be good to share more of these function between the client and the local package but I think this is fine for now.
Contrary to the local package we use `pd.isnull` instead of `check_array(force_finite=True)` because the latter outputs a warning and behave less well with `pd.NA` or None values.

## Breaking changes

---

### DIFF
--- a/tabpfn_client/estimator.py
+++ b/tabpfn_client/estimator.py
@@ -7,6 +7,8 @@ import numpy as np
 import pandas as pd
 from tabpfn_client.config import init
 from sklearn.base import BaseEstimator, ClassifierMixin, RegressorMixin
+from sklearn.utils import column_or_1d
+from sklearn.utils.multiclass import check_classification_targets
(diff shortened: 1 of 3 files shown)


noahho (approved):
Is this consistent with the client? Here, again we have some duplications and a shared interface could help?

LeoGrin:
Mostly consistent, except for this part: "Contrary to the local package we use pd.isnull instead of check_array(force_finite=True) because the latter outputs a warning and behave less well with pd.NA or None values."
Agree once we refactor this should be in a shared package. We could also remove this once errors flow nicely from vertex.

PR #84, opened 2025-02-08
Remaining Credits API

anshulg954:
### Change Description

This is an enhancement where the user will be able to know about his/her used credits and also the total credits. We will also inform the user about the reset time. 

@noahho 

## Breaking changes

NA
---


tabpfn_client/config.py
```diff
@@ -81,3 +81,8 @@ def get_access_token() -> str:
 def set_access_token(access_token: str):
     UserAuthenticationClient.set_token(access_token)
     Config.is_initialized = True
+
+
+def get_usage_details() -> dict:
```
noahho, line 86: similar to server: get_api_usage

tabpfn_client/client.py
```diff
@@ -855,3 +855,23 @@ def try_browser_login(cls) -> tuple[bool, str]:
             return True, token
 
         return False, "Browser login failed or was cancelled"
+
+    @classmethod
+    def get_prediction_hits_data(cls, access_token: str):
```
noahho, line 860: similar to server: get_api_usage

noahho (changes requested):
Great, just improve naming

tabpfn_client/config.py
anshulg954: Done

noahho (approved):

PR #87, opened 2025-02-17
Add credit usage simulation

davidotte:
### Change Description

Add credit usage simulation for TabPFN extensions. For usage just import with `from tabpfn_client.mock_prediction import check_api_credits` and add decorator @check_api_credits to a function. Then, this function will first be simulated (both the execution time and credit usage), it will be checked if the estimated credit usage exceeds the current credits of the user. If yes an error is raised, if not the function gets normally executed.

Few notes:
- With adding a latency offset of 1s for each request, in my tests the simulated cost always overestimated the actual credit usage a bit, so this seems to work quite well. However, I don't know if other users will experience a significantly lower latency.
- For using the decorator in an extension, we probably have to create a common decorator for both client and tabpfn that simply does nothing if tabpfn-client is not used.
- I slightly changed the get_api_usage function Anshul implemented, such that the string for user info is created at config level and not at client level. This allows me to directly access these values.

@noahho 

## Breaking changes

no

no

---


tabpfn_client/estimator.py
```diff
@@ -197,6 +197,8 @@ def _predict(self, X, output_type):
             train_set_uid=self.last_train_set_uid,
             config=estimator_param,
             predict_params={"output_type": output_type},
+            X_train=self.last_train_X,
+            y_train=self.last_train_y,
```
davidotte, line 201: This used to be already there, but it seems like someone removed it, probably accidentally (only for the classifier). This is needed for both the local dataset cache and also for my new things here.

tabpfn_client/mock_prediction.py
```diff
@@ -0,0 +1,202 @@
...
+import warnings
+import logging
+import functools
+from tabpfn_client.client import ServiceClient
+
+
+_thread_local = threading.local()
+
+
+def is_mock_mode() -> bool:
```
noahho, line 16: Add comment for the block of functions: what do they do, why here?

tabpfn_client/mock_prediction.py
```diff
@@ -0,0 +1,202 @@
...
+
+def set_mock_mode(value: bool):
+    setattr(_thread_local, "use_mock", value)
+
+
+def get_cost() -> float:
+    return getattr(_thread_local, "cost", 0.0)
+
+
+def increment_cost(value: float):
```
noahho, line 28: increment_mock_cost

tabpfn_client/mock_prediction.py
```diff
@@ -0,0 +1,202 @@
...
+
+def is_mock_mode() -> bool:
+    return getattr(_thread_local, "use_mock", False)
+
+
+def set_mock_mode(value: bool):
+    setattr(_thread_local, "use_mock", value)
+
+
+def get_cost() -> float:
```
noahho, line 24: get_mock_mode

tabpfn_client/mock_prediction.py
```diff
@@ -0,0 +1,202 @@
...
+    If user has enough credits, function is then executed for real.
+    """
+    from tabpfn_client import get_access_token
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        with mock_mode() as get_cost:
+            func(*args, **kwargs)
+            credit_estimate = get_cost()
+        access_token = get_access_token()
```
noahho, line 187: Do we have code for these two lines already?

tabpfn_client/service_wrapper.py
```diff
@@ -235,12 +240,21 @@ def predict(
...
-            train_set_uid=train_set_uid,
-            x_test=X,
-            tabpfn_config=config,
-            predict_params=predict_params,
-            task=task,
-            X_train=X_train,
-            y_train=y_train,
-        )
+        if is_mock_mode():
+            if X_train is None or y_train is None:
```
noahho, line 244: move the if statement to the mock_predict function?

tabpfn_client/service_wrapper.py
```diff
@@ -235,12 +240,21 @@ def predict(
...
-        return ServiceClient.predict(
-            train_set_uid=train_set_uid,
-            x_test=X,
-            tabpfn_config=config,
-            predict_params=predict_params,
-            task=task,
-            X_train=X_train,
-            y_train=y_train,
-        )
+        if is_mock_mode():
```
noahho, line 243: Write a comment what mock mode is here

noahho (commented):
Great, this looks like a nice solution! Please add some more comments and take a look at my suggestions

tabpfn_client/mock_prediction.py
```diff
@@ -0,0 +1,202 @@
...
+
+def set_mock_time(value: float):
+    setattr(_thread_local, "mock_time", value)
+
+
+def increment_mock_time(seconds: float):
+    set_mock_time(get_mock_time() + seconds)
+
+
+def estimate_duration(
```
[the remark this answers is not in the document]
davidotte: I avoided it so far because it's a bit more work than just copying it to the utils repo. In the server we have different factors we want to get from the subclasses and there the prediction task is also handled as a enum. But I agree that it would be nicer, I will look into it.

tabpfn_client/mock_prediction.py
davidotte: You probably mean get_mock_cost?

tabpfn_client/mock_prediction.py
davidotte: Which two lines exactly do you refer to? That I manually get the access token here? There is already the get_api_usage function in in config.py, but it is currently designed to return an message for the user and not the raw dict. As the duplication is one line of code (getting the access token here), I thought it might be better than defining a seperate function that returns the raw dict.

tabpfn_client/service_wrapper.py
davidotte: But the if statement is for deciding if we want to enter mock_predict in the first place, right? So why do you want to move it into mock_predict?

noahho: I mean "if X_train is None or y_train is None:"

tabpfn_client/mock_prediction.py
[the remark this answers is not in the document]
noahho: ok we spoke about Liam moving and then we have another post-PR

tabpfn_client/service_wrapper.py
davidotte: Ahh! :smile: Yes, makes sense.

noahho:
Great, LGTM!

PR #91, opened 2025-02-27
Httpx ci

### LINKED ISSUE #52: BUG - Understand why the tests are failing with httpx==0.28.1
See #51 and #49 for context. For now we enforce `httpx==0.27.2` but we should be able to use newer version in the future.

mert-kurttutan:
### Change Description

Fix #52 

- No new dependencies

## Breaking changes

- No breaking change

---

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 ### additional requirements for development
 
 # lint and format checking
+pytest
 pre-commit==4.0.1
 ruff == 0.9.1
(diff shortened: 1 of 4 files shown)


mert-kurttutan:
- Resolves #52 . You can also see the reason behind the bug in #52.
- Adds pytest to dev requirements since it is used for testing

LeoGrin:
Thanks for looking into this @mert-kurttutan, LGTM! Could you just add a quick comment in the code to explain the change, and then we can merge :)

PR #97, opened 2025-03-18
replace cityhash with hashlib md5 and upgrade max python to 3.13

LeoGrin:
### Change Description

`python-cityhash` is the only thing preventing us to moving to python3.12 and python3.13 (see https://github.com/escherba/python-cityhash/issues/88). This PR replace this with native hashlib md5 and upgrade max python to 3.13 (and remove max python version all-together, see discussion [here](https://github.com/PriorLabs/TabPFN/issues/234) for the local package).

## Breaking changes

---

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -34,10 +34,10 @@ jobs:
             python-version: "3.9"
             dependency-set: minimum
           - os: ubuntu-latest
-            python-version: "3.12"
+            python-version: "3.13"
             dependency-set: maximum
           - os: macos-latest
-            python-version: "3.12"
+            python-version: "3.13"
             dependency-set: maximum
     runs-on: ${{ matrix.os }}
     
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,10 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "tabpfn-client"
 version = "0.1.6"
-requires-python = ">=3.9,<=3.12"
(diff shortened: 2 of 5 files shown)


davidotte (commented):
Generally looks good. One thing: Hashing with md5 is much slower compared to cityhash. I tried for a dataset with 10.000 rows and 500 columns, there cityhash takes 0.09s and md5 0.28s. Given that predictions for large datasets take a lot of time, this is probably a minor problem. Still, maybe it makes more sense to use a faster hash function or do you think it is fine?

LeoGrin:
Thanks @davidotte :) Knew hashlib would be slower but didn't realise by how much! I've now changed to it `xxhash` which is a dependency but for which the python package seems to be more actively maintained, and the speed seems comparable:
```
Generating random dataset (10,000 x 500)...
Dataset size (MB): 38.14697265625

Serializing dataset...
Serialized size (MB): 91.88809490203857

Benchmarking hash functions (average of 5 runs)...

Results:
------------------------------------------------------------
Hash Function        | Time (seconds)  | Relative Speed 
------------------------------------------------------------
CityHash             | 0.004326s        | 1.00x
xxHash               | 0.005617s        | 1.30x
BLAKE2b              | 0.067008s        | 15.49x
MD5                  | 0.113815s        | 26.31x
Python hash (est.)   | 1.657202s        | 383.08x
------------------------------------------------------------

davidotte (approved):
Thanks! LGTM :)

PR #106, opened 2025-04-12
Add usage analytics tracking

liam-sbhoo:
### Change Description

**Add minimal usage analytics tracking**:
- `tabpfn_client` extracts basic usage info based on the caller's environment, and sends this info in request headers.
- tracking non-privacy intrusive metadata, see "tabpfn_common_utils/usage_analytics/analytics_definition.py"
  ``` python
  ANALYTICS_TO_TRACK = [
    ("X-Unique-Call-Id", get_unique_call_id),
    ("X-Python-Version", get_python_version),
    ("X-Calling-Class", get_calling_class),
    ("X-Module-Name", None),  # Value provided by AnalyticsHttpClient 
  ]
  ``` 

## Breaking changes

---

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -25,6 +25,7 @@ from tabpfn_client.tabpfn_common_utils import utils as common_utils
 from tabpfn_client.constants import CACHE_DIR
 from tabpfn_client.browser_auth import BrowserAuthHandler
 from tabpfn_client.tabpfn_common_utils.utils import Singleton
+from tabpfn_client.tabpfn_common_utils.usage_analytics import AnalyticsHttpClient
 
 logger = logging.getLogger(__name__)
 
@@ -157,10 +158,11 @@ class ServiceClient(Singleton):
     httpx_timeout_s = (
         4 * 5 * 60 + 15  # temporary workaround for slow computation on server side
     )
-    httpx_client = httpx.Client(
+    httpx_client = AnalyticsHttpClient(
         base_url=base_url,
         timeout=httpx_timeout_s,
         headers={"client-version": get_client_version()},
+        module_name="tabpfn_client",
     )
 
     _access_token = None
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -169,6 +169,10 @@ class ServiceClient(Singleton):
     dataset_uid_cache_manager = DatasetUIDCacheManager()
 
     @classmethod
+    def set_module_name(cls, module_name: str):
+        cls.httpx_client.module_name = module_name
+
+    @classmethod
     def get_access_token(cls):
         return cls._access_token
 
--- a/tabpfn_client/config.py
+++ b/tabpfn_client/config.py
@@ -21,7 +21,10 @@ class Config:
     use_server = False
 
 
-def init(use_server=True):
+def init(
+    use_server=True,
+    module_name="tabpfn_client",
+):
     # initialize config
     Config.use_server = use_server
 
@@ -30,6 +33,8 @@ def init(use_server=True):
         return
 
     if use_server:
+        ServiceClient.set_module_name(module_name)
+
         # check connection to server
         if not UserAuthenticationClient.is_accessible_connection():
             raise RuntimeError(


tabpfn_client/client.py
```diff
@@ -157,15 +158,20 @@ class ServiceClient(Singleton):
...
         timeout=httpx_timeout_s,
         headers={"client-version": get_client_version()},
+        module_name="tabpfn_client",
     )
 
     _access_token = None
     dataset_uid_cache_manager = DatasetUIDCacheManager()
 
+    @classmethod
+    def set_module_name(cls, module_name: str):
```
Jabb0, line 172: Please add return type annotations.

tabpfn_client/client.py
```diff
@@ -157,15 +158,20 @@ class ServiceClient(Singleton):
     httpx_timeout_s = (
         4 * 5 * 60 + 15  # temporary workaround for slow computation on server side
     )
-    httpx_client = httpx.Client(
+    httpx_client = AnalyticsHttpClient(
         base_url=base_url,
         timeout=httpx_timeout_s,
         headers={"client-version": get_client_version()},
+        module_name="tabpfn_client",
```
Jabb0, line 165: Are you sure there is no way of getting the module name programmatically?

tabpfn_client/client.py
```diff
@@ -157,15 +158,20 @@ class ServiceClient(Singleton):
     httpx_timeout_s = (
         4 * 5 * 60 + 15  # temporary workaround for slow computation on server side
     )
-    httpx_client = httpx.Client(
+    httpx_client = AnalyticsHttpClient(
```
Jabb0, line 161: Why is this not in the constructor?

tabpfn_client/config.py
```diff
@@ -30,6 +33,8 @@ def init(use_server=True):
         return
 
     if use_server:
+        ServiceClient.set_module_name(module_name)
```
Jabb0, line 36: You are not expecting this to change during service run. It should be ONLY in the constructor of ServiceClient and then stored as a private member.

tabpfn_client/client.py
```diff
@@ -157,15 +158,20 @@ class ServiceClient(Singleton):
...
         base_url=base_url,
         timeout=httpx_timeout_s,
         headers={"client-version": get_client_version()},
+        module_name="tabpfn_client",
     )
 
     _access_token = None
     dataset_uid_cache_manager = DatasetUIDCacheManager()
```
Jabb0, line 169: should be private btw.

tabpfn_client/client.py
```diff
@@ -157,15 +158,20 @@ class ServiceClient(Singleton):
     httpx_timeout_s = (
         4 * 5 * 60 + 15  # temporary workaround for slow computation on server side
     )
-    httpx_client = httpx.Client(
```
Jabb0, line 160: Tests missing.

Jabb0 (changes requested):

tabpfn_client/client.py
liam-sbhoo: At least not robustly, in my opinion.
My rationale: `tabpfn-extensions` or wrappers built by users can create arbitrarily deep call stack. So I thought it'd be easier to let the wrapper, e.g. `tabpfn-extensions`, set the module name explicitly.

tabpfn_client/client.py
liam-sbhoo: This singleton class doesn't have a constructor.

tabpfn_client/config.py
liam-sbhoo: ServiceClient is a singleton class, hence we don't have an constructor as per say.
I'm also wondering if there's a better design for classes like `ServiceClient`

tabpfn_client/client.py
Jabb0: I see. I was hoping there is a better way to get the module name than iterating the call stack.

tabpfn_client/config.py
Jabb0: There should be. Singletons are to be handled with cautions, because as for global variables, they can be changed in multiple places and thus introduce side-effects. Also they are harder to test, because dependency injection does not work.

I'd need to think more about this here.

But one way would be to create a client instance. And then this instance needs to be passed when calls are made.

PR #120, opened 2025-07-22
Updated user fields

Jabb0:
### Change Description

- Made role, company and usecase mandatory fields.

- No

## Breaking changes

---


requirements.txt
```diff
@@ -6,3 +6,4 @@ password-strength>=0.0.3.post2,<=0.0.3.post2
 sseclient-py>=1.8.0,<=1.8.0
 tqdm>=4.63.0,<=4.67.1
 xxhash>=1.1.0,<=3.5
+inquirer>=3.4.0
```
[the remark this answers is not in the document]
brendan-priorlabs: This is pretty suspect advice IMO. Lockfiles are the way to handle this. Otherwise when we need to upgrade something it won't be clear which of the maximum pins are important and which were added speculatively.

requirements.txt
```diff
@@ -6,3 +6,4 @@ password-strength>=0.0.3.post2,<=0.0.3.post2
 sseclient-py>=1.8.0,<=1.8.0
 tqdm>=4.63.0,<=4.67.1
 xxhash>=1.1.0,<=3.5
+inquirer>=3.4.0,<4.0.0
```
noahho: how did you come to choose this package vs e.g. ruff or others?

requirements.txt
```diff
@@ -6,3 +6,4 @@ password-strength>=0.0.3.post2,<=0.0.3.post2
 sseclient-py>=1.8.0,<=1.8.0
 tqdm>=4.63.0,<=4.67.1
 xxhash>=1.1.0,<=3.5
+inquirer>=3.4.0,<=4.0.0
```
brendan-priorlabs: Looks like this is failing because [4.0.0 doesn't exist yet](https://github.com/PriorLabs/tabpfn-client/actions/runs/16469283960/job/46554214759?pr=120#step:8:91). I'd just pick the highest version that exists and leave a TODO in the requirements.txt for us to fix that at some point. Possibly/hopefully that'll get obviated by the `uv` change when we get to it

brendan-priorlabs: Actually I guess all the deps suffer from this, so no need for a TODO 🙃

requirements.txt
Jabb0: you mean rich?

because it does not have a dropdown.

requirements.txt
Jabb0: the highest version is 3.4.0

Jabb0: added todo and set to this version.

requirements.txt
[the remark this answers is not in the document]
Jabb0: As discussed I think this advise follows the SemVer convention and assumes major versions are to be inspected, while minor and patch versions are fine.

As you said SemVer is a convention and this must not hold true. Therefore, it's not a guarantee and using it a tradeoff with larger version compatibility.

As discussed: Remove the maximum version for libraries and use dependency locking for applications.

tabpfn_client/prompt_agent.py
```diff
@@ -251,15 +253,51 @@
...
+                "\nPlease help us tailor our support and services to your needs"
             )
             + "\n"
         )
 
-        company = input(cls.indent("Where do you work? "))
-        role = input(cls.indent("What is your role? "))
-        use_case = input(cls.indent("What do you want to use TabPFN for? "))
+        while True:
+            company = input(cls.indent("Where do you work? "))
```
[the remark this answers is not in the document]
Jabb0: done

tabpfn_client/prompt_agent.py
```diff
@@ -251,15 +253,51 @@
...
+                    message="Please specify your role",
+                    validate=lambda _, x: re.match(r'\S+', x),
+                )
+            ]
+            role = inquirer.prompt(other_question)["other_role"]
+        else:
+            role = answers["role"]
+
+        while True:
+            use_case = input(cls.indent("What do you want to use TabPFN for? "))
```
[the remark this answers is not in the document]
Jabb0: done

.github/workflows/pull_request.yml
```diff
@@ -72,6 +72,9 @@ jobs:
               if match:
                   package, min_ver, _ = match.groups()
                   min_reqs.append(f"{package}=={min_ver}")
+              else:
+                print(f"ERROR: Unmatched dependency found: {req}")
+                exit(1)
```
brendan-priorlabs, line 77: Just a tip, in general I like raising instead of calling exit so that there's a stack trace. Doesn't matter much in this yml config -- simply an idea for next time.

brendan-priorlabs (approved):
LGTM. Thanks, Felix!

.github/workflows/pull_request.yml
Jabb0: got it. thanks!

PR #122, opened 2025-07-28
Minor fixes for 0.1.9 release

brendan-priorlabs:
- Bump version
- Add twine to README.md for release process
- Move typing_extensions from dev to main `requirements.txt`. This seems necessary. Otherwise we have failures running `quick_test.py` which is really just a simple demo of the package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "tabpfn-client"
-version = "0.1.9"
+version = "0.1.8"
 requires-python = ">=3.9"
 dynamic = ["dependencies", "optional-dependencies"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "tabpfn-client"
-version = "1.8.1"
+version = "0.1.9"
 requires-python = ">=3.9"
 dynamic = ["dependencies", "optional-dependencies"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "tabpfn-client"
-version = "0.1.7"
+version = "1.8.1"
 requires-python = ">=3.9"
 dynamic = ["dependencies", "optional-dependencies"]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "tabpfn-client"
-version = "0.1.7"
+version = "0.1.9"
```
brendan-priorlabs, line 7: Not 0.1.8 because I already uploaded that to the test pypi

LeoGrin: I think you can have the same version on pypi and on test pypi?

LeoGrin (approved):
LGTM once the tests pass, just one comment on the version number.
Also we have a script to extract the min / max dependencies of the package, which fails when you pin an exact version. A slightly hacky fix is to use somethign like `sseclient-py>=1.8.0,<=1.8.0`, otherwise the script could be improved.

pyproject.toml
brendan-priorlabs: Can and should IMO. The problem is that 0.1.8 is broken. That's the problem this PR is trying to fix. I can upload it as 1.9.1rc1 or something on the test pypi initially so that we I don't have to randomly bump the number again if it still doesn't work.

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 ### additional requirements for development
 
 # lint and format checking
-pytest
 pre-commit==4.2.0
+pytest
```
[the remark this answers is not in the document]
brendan-priorlabs: I'm somewhat skeptical of this (seems like we're misusing requirements.txt as a lockfile), but going with it for now.

pyproject.toml
LeoGrin: Ah but IIRC you can redeploy a version in test pypi, which would replace 0.1.8 by the new one. But if you think it's simpler to bump 0.1.9 directly, fine for me!

brendan-priorlabs: Note: We tried this (even deleting the bad 0.1.8 on test pypi), but it doesn't seem to work. Version numbers really can't be shared.

brendan-priorlabs:
Thanks for the review, @LeoGrin!

PR #144, opened 2025-10-22
Allow 50k rows datasets w/ uploads over http2

simo-prior:


safaricd (approved):
All looking good!

tabpfn_client/server_config.yaml
```diff
@@ -1,8 +1,18 @@
...
+# preprod
+# protocol: "https"
+# host: "tabpfn-server-preprod-1039906307296.europe-west4.run.app"
+# port: "443"
+
+# development
+# protocol: "https"
+# host: "api-813747016009.europe-west4.run.app"
+# port: "443"
+
```
brendan-priorlabs: Not that these are truly secret, but we should not include them in the public client

```suggestion
```

tabpfn_client/estimator.py
```diff
@@ -16,7 +16,7 @@
 
 logger = logging.getLogger(__name__)
 
-MAX_ROWS = 10000
+MAX_ROWS = 50_000
 MAX_COLS = 500
 MAX_NUMBER_OF_CLASSES = 10
```
noahho, line 21: We should introduce a number of cells limit too. 50K x 500 is too large and we could start with handling data up to 50K x 200 features or 20K x 500 features, i.e. max cells = 50_000 * 200

simo-prior: Changing to 400 :+1

PR #145, opened 2025-10-25
Refactor: Overhaul Authentication & Registration UI/UX

noahho:
### Description

This PR introduces a significant overhaul of the client's interactive authentication and registration flow. The primary goal is to improve the user experience by replacing the old, text-heavy `print` statements with a modern, interactive, and more guided process using the `rich` library.

Two things that might need discussing on highlevel:
- Rich is added for better formatting, it’s a lightweight library, pretty common but it’s a dependency
- I added a Registration Progress Saver, that reloads half finished registration flow if a user aborts in the middle. That’s not essential and a bit more complex.

This new flow is more robust, handling edge cases like interrupted registrations, unverified emails, and internet connectivity issues, while providing clearer feedback to the user at every step.

image

* * *

### Key Changes

-   **🎨 New UI Module & `rich` Dependency:**
    
    -   Adds `rich` as a new dependency.
        
    -   Introduces a new `tabpfn_client/ui.py` module to centralize all styled terminal output (e.g., `success`, `warn`, `fail`, `status`, `header`, and ASCII logos).
        
    -   All interactive prompts in `prompt_agent.py` are refactored to use these new UI components instead of `print` and manual ANSI color codes.
        
-   **⚙️ Improved Registration Flow:**
    
    -   The registration process is now a guided, multi-step flow (e.g., "Step 1/6...").
        
    -   **Registration State Management:** A new `RegistrationState` manager (`tabpfn_client/state_manager.py`) saves progress to `~/.tabpfn/registration_state.json`. If registration is interrupted (e.g., `Ctrl+C`), the user is prompted to resume from where they left off.
        
    -   **Better Password Validation:** As the user types a password, the UI now interactively shows which requirements are met and which are not.
        
    -   **Simplified Prompts:** Prompts for Terms & Conditions and Data Privacy are clearer and more concise.
        
-   **🔒 Enhanced Login & Verification:**
    
    -   **Internet Check:** The client now checks for an internet connection before attempting any network requests.
        
    -   **Improved Login Failure:** When a login fails, the user is given clear options to "Try password again," "Reset password," or "Change email."
        
    -   **Interactive Email Verification:** The email verification prompt now accepts commands:
        
        -   `resend`: To request a new verification code.
            
        -   `quit`: To exit the verification process.
            
    -   **Unverified Email Handling:** If a user logs in but their email is not yet verified, they are guided to the verification step instead of just failing.
        
-   **🧹 Cleanup & Testing:**
    
    -   Removes unused `os.system("clear")` calls, replacing them with `console.clear()`.
        
    -   Removes the `dependencies` list from `pyproject.toml` (as they are managed in `requirements.txt`).
        
    -   Updates all unit tests (especially `test_prompt_agent.py`) to mock the new `rich.console.Console.input` and align with the new interactive flows.

### DIFF
--- a/tabpfn_client/ui.py
+++ b/tabpfn_client/ui.py
@@ -128,14 +128,14 @@ _PRIOR_LABS_ASCII_SMALL = r"""
 """
 
 
-def print_logo(subtitle = None) -> None:
+def print_logo(subtitle=None) -> None:
     """Print the large Prior Labs ASCII logo with optional subtitle."""
     console.print(_PRIOR_LABS_ASCII, style="bold blue")
     if subtitle:
         console.print(f"[dim]{subtitle}[/dim]", end="\n\n")
 
 
-def print_logo_small(subtitle = None) -> None:
+def print_logo_small(subtitle=None) -> None:
     """Print a small Prior Labs ASCII banner with optional subtitle."""
     console.print(_PRIOR_LABS_ASCII_SMALL, style="bold blue")
     if subtitle:


pyproject.toml
```diff
@@ -27,9 +27,6 @@ classifiers = [
   'Programming Language :: Python :: 3.13',
 ]
 license = { file = "LICENSE" }
-dependencies = [
```
noahho: I removed here as install didn't work out of the box. This dependency conflicted with the dependencies specified in requirements below.

brendan-priorlabs: This is critical though. We need http2 for large files

noahho: It is placed in the requirements file where all other deps live. For me having deps in both places caused an install error with pip.

noahho: In requirements: httpx[http2]>=0.25.0,<=0.28.1

brendan-priorlabs: Ah, gotcha!

tabpfn_client/prompt_agent.py
```diff
@@ -35,290 +46,502 @@ def password_req_to_policy(password_req: list[str]):
...
+                warn("Please enter 'y' or 'n'.")
+
+    @classmethod
+    def prompt_personally_identifiable_information_simple(cls) -> bool:
+        """Simplified data privacy prompt for registration flow."""
+        console.print("Do not upload personal/sensitive data.")
+
+        while True:
+            choice = (
+                console.input("[bold blue]→[/bold blue] I understand (y/n): ")
```
brendan-priorlabs: I am not a lawyer, but I think "I agree" might be better here

tabpfn_client/prompt_agent.py
```diff
@@ -35,290 +46,502 @@ def password_req_to_policy(password_req: list[str]):
...
+            )
+            if choice in ["y", "yes"]:
+                return True
+            elif choice in ["n", "no"]:
+                return False
+            else:
+                warn("Please enter 'y' or 'n'.")
+
+    @classmethod
+    def prompt_personally_identifiable_information_simple(cls) -> bool:
```
brendan-priorlabs: This simple/non-simple distinction is kind of surprising. Do we really want both? Seem ripe for drift

tabpfn_client/service_wrapper.py
```diff
@@ -142,6 +143,18 @@ def verify_email(cls, token: str, access_token: str) -> tuple[bool, str]:
     @classmethod
     def try_browser_login(cls) -> tuple[bool, str]:
         """Try to authenticate using browser-based login"""
+
+        # Allow disabling browser login via environment variables
+        # - TABPFN_NO_BROWSER=1/true/yes/on
+        # - TABPFN_CLIENT_NO_BROWSER=1/true/yes/on (backward-compatible alias)
```
brendan-priorlabs, line 149: This is new right? Nothing to be backward compatible with

https://github.com/search?q=repo%3APriorLabs%2Ftabpfn-client%20TABPFN_CLIENT_NO_BROWSER&type=code

noahho: yes it's new

PR #147, opened 2025-10-27
Default model selection for client

brendan-priorlabs:
Stop trying to be clever about the default model. Let the server decide.


tabpfn_client/estimator.py
```diff
@@ -45,8 +45,9 @@ def _model_name_to_path(
     ) -> str:
```
klemens-floege:
```suggestion
    ) -> str | None:
```

PR #159, opened 2025-10-29
Add call metadata

safaricd:
### Change Description

This pull request introduces changes to the APIs `/predict` response for returning metadata like TabPFN model version that was used for inference.

Users can access the property using:

```python
reg = TabPFNRegressor()
reg.predict(....)

print(f"Model metadata: {reg.last_meta}")
```

### DIFF
--- a/tabpfn_client/estimator.py
+++ b/tabpfn_client/estimator.py
@@ -117,7 +117,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator, TabPFNModelSelection):
         random_state: int or RandomState or RandomGenerator or None, default=None
             Controls the randomness of the model. Pass an int for reproducible results.
         inference_config: dict or None, default=None
-            Additional advanced arguments for model interface. See the doc of ModelInterfaceConfig
+            Additional advanced arguments for model interface. See the doc of InferenceConfig
             in the tabpfn package for more details. For the client, the inference_config and the
             preprocess transforms need to be dictionaries.
         paper_version: bool, default=False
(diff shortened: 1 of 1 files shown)


tabpfn_client/client.py
```diff
@@ -391,6 +396,8 @@ def run_progress():
                             progress_thread.start()
                         elif data["event"] == "result":
                             results = data["data"]
+                            # Extract metadata from the response
+                            cls._last_meta = data["metadata"]
```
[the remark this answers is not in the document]
safaricd: Fixed ✅

tabpfn_client/client.py
```diff
@@ -409,6 +396,8 @@ def run_progress():
                             progress_thread.start()
                         elif data["event"] == "result":
                             results = data["data"]
+                            # Extract metadata from the response
+                            cls._last_meta = data.get("metadata", {})
```
brendan-priorlabs: Please make this thread-safe by returning `_last_meta` so we can set it on the estimator instead setting it on the class. Possibly a good idea to make a `predict_with_meta` or similar so that we can preserve the existing interface

safaricd: Good call - had to change the interface and quick a few unit tests with that change but all good now. RE: `predict_with_meta`, I don't see why we should introduce yet an another interface at the moment so let's keep it minimalistic as is.

safaricd: This was addressed by changing the `ServiceClient.predict` signature to return `PredictionResult`. This `dataclasses` wraps the API prediction and metadata - alternatively, we could have subclass `np.ndarray` internally (up until the estimator) to avoid changing unit tests, however I am personally not in favor of subclassing it - e.g. we'd need to re-implement relevant functions like `__array_func__`, `__array_attributes__` etc., and even with that, there might be edge cases where the state if not copied as expected.

brendan-priorlabs: +1 to not subclassing here. As for `predict_with_meta`, the rationale is that a) you wouldn't need to change the unit tests (moot point now) and b) I'm not sure if it's clear that `ServiceClient.predict` is fully internal. If you're convinced it is, SGTM

brendan-priorlabs:
Thanks, Dominik. Feel free to merge once @noahho approves since I'll be out this afternoon

PR #166, opened 2025-11-04
Handle training set as torch tensor

simo-prior:
### Description

https://linear.app/priorlabs/issue/OPC-2/client-fails-on-tensor-data

### DIFF
--- a/tabpfn_client/client.py
+++ b/tabpfn_client/client.py
@@ -749,11 +749,14 @@ class ServiceClient(Singleton):
         if response.status_code == 200:
             access_token = response.json()["access_token"]
             message = ""
-        elif response.status_code == 403:
-            access_token = response.headers["access_token"]
-            message = response.json()["detail"]
         else:
-            message = response.json()["detail"]
+            try:
+                message = response.json()["detail"]
+            except (json.JSONDecodeError, KeyError):
+                message = (
+                    response.text
+                    or f"Login failed with status code {response.status_code}"
+                )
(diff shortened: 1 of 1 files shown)


tabpfn_client/estimator.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import pandas as pd
 from tabpfn_client.config import init
+from torch import Tensor
```
noahho: this needs to be a gated import as we want this to be optional

noahho: try:
    from torch import Tensor
    TORCH_AVAILABLE = True
except ImportError:
    TORCH_AVAILABLE = False

noahho:
```
elif TORCH_AVAILABLE and isinstance(X, Tensor):
        # The code from here on will only run if torch is installed
        # and X is actually a torch Tensor.
        X_tensor = X
        if X_tensor.requires_grad:
            X_tensor = X_tensor.detach()
        if X_tensor.is_cuda:
            X_tensor = X_tensor.cpu()
        
        X_np = X_tensor.numpy()
```

noahho: We need to move to cpu as well if data was on GPU before

simo-prior: @noahho adjusted. I'm assigning back to X so that the current logic for numpy arrays applies, let me know if there are reasons to preserve X's initial reference.

noahho (approved):

PR #184, opened 2025-11-19
WIP: Save/load models for CAAFE -- Client

brendan-priorlabs:


src/tabpfn_client/client.py
```diff
@@ -509,7 +516,7 @@ def fit(
             raise RuntimeError("Error during fit. No valid model received.")
 
         cls.dataset_uid_cache_manager.add_dataset_uid(dataset_hash, train_set_uid)
-        return train_set_uid
+        return FittedModel(train_set_uid=train_set_uid, model_id=model_id)
```
safaricd, line 519: Won't this break backwards compatibility for anyone using the `train_set_uid`?

src/tabpfn_client/client.py
```diff
@@ -664,13 +674,14 @@ def run_progress():
                         raise NotImplementedError(
                             "Automatically re-uploading the train set is not supported for thinking mode. Please call fit()."
                         )
-                    train_set_uid = cls.fit(
+                    fitted_model = cls.fit(
```
safaricd, line 677: Could you please share more context on why do we re-fit it again in `predict`?

PR #194, opened 2025-12-09
Trigger multi-gpu mode in quick_test.py

brendan-priorlabs:


tests/quick_test.py
```diff
@@ -18,18 +19,27 @@
...
+        big = pd.concat([df] * 50, ignore_index=True)
+        print(f"shape after: {big.shape}")
+        return big
 
 if __name__ == "__main__":
     # Patch webbrowser.open to prevent browser login
     with patch("webbrowser.open", return_value=False):
-        use_server = True
-        # use_server = False
+        trigger_multi_gpu_threshold = True
```
safaricd, line 32: We will always trigger a multi GPU setup. Can we change this such that we test both - single & multi GPU?

PR #196, opened 2025-12-12
Set token before connection probing

simo-prior:
### Description

Internal environments will require passing the token for all the endpoints due to auth proxies.

### DIFF
--- a/src/tabpfn_client/config.py
+++ b/src/tabpfn_client/config.py
@@ -38,19 +38,16 @@ def init(use_server=True):
         return
 
     if use_server:
-        is_valid_token, access_token = (
-            UserAuthenticationClient.try_reuse_existing_token()
-        )
-        # try:
-        #     is_valid_token, access_token = (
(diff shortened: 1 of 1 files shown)


src/tabpfn_client/config.py
```diff
@@ -31,17 +31,14 @@ def init(use_server=True):
         return
 
     if use_server:
+        is_valid_token, access_token = UserAuthenticationClient.try_reuse_existing_token()
```
brendan-priorlabs: Can you determine whether the error we get back is a 403 vs. just no connection so we can preserve the user-friendly error below?

brendan-priorlabs (approved):
Approving now in case this is urgent, but please do consider that comment for user-friendliness!

src/tabpfn_client/config.py
simo-prior: @brendan-priorlabs addressed

PR #203, opened 2026-01-01
Document public user login functions

noahho:
Added detailed docstrings for client functions to improve documentation and usability.

### Change Description

## Breaking changes

---

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -99,6 +99,15 @@ jobs:
       - name: Initialize submodules
         run: git submodule update --init --recursive
 
+      - name: Check for forbidden licenses
+        run: |
+          licensecheck \
+            --requirements-paths pyproject.toml \
+            --only-licenses APACHE MIT BSD ISC PYTHON UNLICENSE UNKNOWN \
+            --ignore-packages certifi "tabpfn*" \
+            --show-only-failing \
+            -0
+
       - name: Report cache status
         run: |
           echo "Cache status: ${{ steps.pip-cache.outputs.cache-hit == 'true' && 'HIT' || 'MISS'}}"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "tabpfn-client"
-version = "0.2.9rc5"
+version = "0.2.9rc6"
 requires-python = ">=3.9"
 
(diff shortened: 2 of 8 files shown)


src/tabpfn_client/config.py
```diff
@@ -30,6 +30,17 @@ def __new__(cls, *args, **kwargs):
 
 
 def init(use_server=True):
+    """
+    Initializes the TabPFN client and handles user authentication.
```
safaricd, line 34: Nicely explained! Just one small addition I'd personally add is noting that this is mostly an internal method, used in e.g. `get_access_key`. We don't advertise for using `init` in the official docs, so just to make it clear for the users what the usual & official interface is.

src/tabpfn_client/config.py
```diff
@@ -101,16 +118,42 @@ def reset():
...
+def get_api_usage() -> str:
+    """
+    Fetches and formats the current API usage statistics for the user.
+
+    :return: A human-readable string detailing current credit usage, 
+             the total limit, and when the limit resets.
+    """
     access_token = get_access_token()
     response = ServiceClient.get_api_usage(access_token)
     return f"Currently, you have used {response['current_usage']} of the allowed limit of {'Unlimited' if int(response['usage_limit']) == -1 else response['usage_limit']} credits. The limit will reset at {response['reset_time']}."
```
safaricd, line 159: I know that this is out of scope of the current PR (which focuses on documentation), but I'd rather have the `get_api_usage` function return a `dict` or `dataclass` like `TabPFNQuota`:
```python
from dataclasses import dataclass

@dataclass
class TabPFNQuota:
    current_usage: int
    usage_limit: int
    reset_time: datetime
```
This will make the quota reusable within code; otherwise, users will have to manually parse the string using regex e.g. The `dataclass` might also implement the `__repr__` interface, which would then essentially implement the current behavior. WDYT?

noahho: absolutely agree

src/tabpfn_client/config.py
noahho: I think we use this function in some places to trigger the login, so while we don't advertise it in docs, it's used externally

brendan-priorlabs (approved):

src/tabpfn_client/config.py
noahho: I'll create a ticket

PR #239, opened 2026-04-12
Add tests for v2 & handle deprecation

simo-prior:
- [x] add v2 tests
- [x] fix bug w/ `Prediction` having too narrow type
- [x] add mechanism to handle deprecation headers

### DIFF
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -10,7 +10,7 @@ Prediction = Union[
     List[PredictionScalar],
     List[float],
     List[List[float]],
-    Dict[str, Union[List[float], List[List[float]]]],
+    Dict[str, Union[List[Optional[float]], List[List[Optional[float]]]]],
 ]
 
 TabPFNConfig = Optional[Dict[str, Any]]
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -55,10 +55,13 @@ class FileUploadInfo(BaseModel):
 # /tabpfn/get_dataset_limits/
 # ---------------------------------------------------------------------------
 class GetDatasetLimitsResponse(BaseModel):
-    max_size_bytes: int
-    max_cells: int
-    max_cols: int
-    max_classes: int
+    dataset_max_size_bytes: int
+    dataset_max_cols: int
+    dataset_max_classes: int
+    train_set_max_rows: int
+    train_set_max_cells: int
+    test_set_max_rows: int
+    test_set_max_rows_w_full_regression_output: int
 
 
 # ---------------------------------------------------------------------------
--- a/src/tabpfn_client/client.py
+++ b/src/tabpfn_client/client.py
@@ -211,7 +211,7 @@ class ServiceClient(Singleton):
     httpx_client = httpx.Client(
         base_url=base_url,
         timeout=TABPFN_CLIENT_TIMEOUT,
-        headers={"client-version": get_client_version()},
+        headers={"Prior-Client-Version": get_client_version()},
         transport=SelectiveHTTP2Transport(http2_paths=[fit_path]),
         follow_redirects=True,
     )
@@ -1128,7 +1128,7 @@ class ServiceClient(Singleton):
             full_url,
             headers={
                 "Authorization": f"Bearer {cls.get_access_token()}",
-                "client-version": get_client_version(),
+                "Prior-Client-Version": get_client_version(),
             },
         ) as response:
             cls._validate_response(response, "download_all_data")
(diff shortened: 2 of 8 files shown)


src/tabpfn_client/client.py
```diff
@@ -1128,7 +1159,7 @@ def download_all_data(cls, save_dir: Path) -> Union[Path, None]:
             full_url,
             headers={
                 "Authorization": f"Bearer {cls.get_access_token()}",
-                "client-version": get_client_version(),
+                "Prior-Client-Version": get_client_version(),
```
ggprior, line 1162: minor: what's the reasoning behind changing this header?

src/tabpfn_client/client.py
```diff
@@ -687,6 +697,25 @@ def _raise_not_found_error(error_response: ErrorResponse) -> None:
             raise NeedsRefittingError(error_response.message)
         raise RuntimeError(error_response.message)
 
+    @staticmethod
+    def _warn_if_deprecated(response: httpx.Response) -> None:
```
ggprior, line 701: nit: will likely result in two separate warnings for now, one through greeting (though guessing that's temporary), one through this more long-term approach. But I think it's fine.

src/tabpfn_client/api_models.py
```diff
@@ -55,10 +55,13 @@ class FileUploadInfo(BaseModel):
 # /tabpfn/get_dataset_limits/
 # ---------------------------------------------------------------------------
 class GetDatasetLimitsResponse(BaseModel):
-    max_size_bytes: int
-    max_cells: int
-    max_cols: int
-    max_classes: int
+    dataset_max_size_bytes: int
+    dataset_max_cols: int
+    dataset_max_classes: int
```
ggprior, line 60: are the old fields still returned to old clients for backwards compat?

ggprior (approved):
LGTM on the deprecation handling, I'll leave the other reviewers to comment on the limits.

src/tabpfn_client/client.py
simo-prior: `client-version` is quite generic, better to namespace headers to avoid conflicts.

src/tabpfn_client/client.py
simo-prior: @ggprior This warning is rather for future deprecations: users on the latest client won't see the warning cause they don't need to upgrade and those who do need to upgrade won't have this code in their client.

src/tabpfn_client/api_models.py
simo-prior: This endpoint has not been released yet officially, we can still afford to break things :)

src/tabpfn_client/api_models.py
```diff
@@ -55,10 +55,13 @@ class FileUploadInfo(BaseModel):
 # /tabpfn/get_dataset_limits/
 # ---------------------------------------------------------------------------
 class GetDatasetLimitsResponse(BaseModel):
```
noahho, line 57: get_dataset_limits() is a single global call with no model version parameter - are the limits actually the same across v2/v2.5/v2.6? If they differ per model, we'd want to pass the version through so client-side validation doesn't incorrectly reject or allow requests?

noahho, line 57: get_dataset_limits() is a single global call with no model version parameter - limits are not the same across v2/v2.5/v2.6 so, we'd want to pass the version through so client-side validation doesn't incorrectly reject or allow requests?

src/tabpfn_client/estimator.py
```diff
@@ -639,14 +633,40 @@ def validate_data_size(X: np.ndarray, y: Union[np.ndarray, None] = None):
...
-            f"The number of columns ({X.shape[1]}) exceeds the maximum of {limits.max_cols}."
+            f"The number of rows ({X.shape[0]}) exceeds the maximum of {limits.train_set_max_rows}."
         )
+    if X.shape[1] > limits.dataset_max_cols:
+        raise ValueError(
+            f"The number of columns ({X.shape[1]}) exceeds the maximum of {limits.dataset_max_cols}."
+        )
+
+
+def validate_test_set(X: np.ndarray, output_type: str):
```
noahho, line 650: We will want to validate the number of text cells as well separately. This could happen within that text module though. Maybe makes sense to do validation in one place though and call some kind of pipeline validation?

noahho (approved):

src/tabpfn_client/api_models.py
simo-prior: @noahho I think in the future we can document limits per model and for old clients that do not implement a per-model logic we can send down the last support model limits by inspecting the client-version header.

src/tabpfn_client/estimator.py
simo-prior: Added the check on max cells for the test set 👍

PR #243, opened 2026-04-16
Ensure data max size limit applies to in-memory size

simo-prior:


src/tabpfn_client/client.py
```diff
@@ -310,18 +309,22 @@ def fit(
 
         client_options = client_options or ClientOptions()
 
-        x_bytes, x_crc32c_hash = _serialize_to_parquet(X)
-        y_bytes, y_crc32c_hash = _serialize_to_parquet(y)
+        df_X = X if isinstance(X, pd.DataFrame) else pd.DataFrame(X)
```
safaricd, line 312: Casting `pd.DataFrame` with `X` which is untyped without any safety checks may be too open?

src/tabpfn_client/client.py
```diff
@@ -145,13 +145,12 @@ def _get_crc32c_hash(data: bytes) -> str:
...
-    df = data if isinstance(data, pd.DataFrame) else pd.DataFrame(data)
+def _serialize_to_parquet(df: pd.DataFrame) -> tuple[bytes, str]:
     buf = io.BytesIO()
     df.to_parquet(buf, index=False, compression="zstd")
-    dataset_bytes = buf.getvalue()
-    crc32c_b64 = _get_crc32c_hash(dataset_bytes)
-    return dataset_bytes, crc32c_b64
+    parquet_bytes = buf.getvalue()
+    crc32c_b64 = _get_crc32c_hash(parquet_bytes)
+    return parquet_bytes, crc32c_b64
```
safaricd: Any perf concerns here? `buf.getvalue()` copies the serialized bytes, so peak memory briefly hits ~2× the Parquet payload.

brendan-priorlabs (approved):
LGTM!

src/tabpfn_client/client.py
simo-prior: @safaricd added types, the upstream functions were already typed but this one wasn't 👍

src/tabpfn_client/client.py
simo-prior: Made a small change to improve things, but the memory size of the dataframe can be much bigger than 2 copies of compressed parquet anyway.

PR #252, opened 2026-04-22
Version dataset limits by model

simo-prior:


safaricd (approved):

src/tabpfn_client/constants.py
```diff
@@ -1,20 +1,42 @@
...
 logger = logging.getLogger(__name__)
 
 
+# TODO: This should be exported by a tabpfn-types package at some point.
 class ModelVersion(str, Enum):
-    """Version of the model."""
-
     V2 = "v2"
     V2_5 = "v2.5"
+    V2_6 = "v2.6"
```
ggprior: not sure we'll ship 2.6 through API

ggprior: in current setup, definitely wouldn't work

ggprior (approved):
looks good! one comment on 2.6
Question -- will this already apply the updated v3 limits? (notably 160 classes). If so, needs to go in ASAP as currently the limit for v3 is 10 classes still.

PR #263, opened 2026-04-30
Adjust interfaces pre-release

simo-prior:
- [x] http2 on `/tabpfn/predict` for forward compatibility.
- [x] move client_options to the estimator interface.
- [x] add `force_refit` as an option to the estimator interface.
- [x] adjust way we create a sentry-trace as headers are set on the estimator now instead of the method.

### DIFF
--- a/src/tabpfn_client/client.py
+++ b/src/tabpfn_client/client.py
@@ -274,8 +274,8 @@ class ServiceClient(Singleton):
         task: Literal["classification", "regression"],
         tabpfn_config: Union[dict, None] = None,
         description: str | None = None,
+        force_refit: bool = False,
         client_options: ClientOptions | None = None,
-        is_refitting: bool = False,
     ) -> UUID:
         """
         Upload a train set to server and return the train set UID if successful.
@@ -293,6 +293,8 @@ class ServiceClient(Singleton):
             `paper_version`.
         description: str, optional
             Description of the dataset and task for the server.
+        force_refit: bool, optional
+            Whether to force refit the model even if the model has already been fitted.
         client_options : ClientOptions, optional
             Per-request options (e.g. timeout, headers) for the fitting API call
             only. Does not apply to file uploads. Because uploads can run before fitting,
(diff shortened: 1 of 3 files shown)


src/tabpfn_client/estimator.py
```diff
@@ -647,7 +633,7 @@ def predict_task() -> PredictionResult:
                         task="regression",
                         tabpfn_config=estimator_param,
                         description=self.last_train_set_description,
-                        client_options=client_options,
+                        client_options=self.client_options,
                         is_refitting=True,
                     )
```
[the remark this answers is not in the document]
simo-prior: Hallucination

src/tabpfn_client/client.py
```diff
@@ -216,11 +216,12 @@ class ServiceClient(Singleton):
         or f"{server_config.protocol}://{server_config.host}:{server_config.port}"
     )
     fit_path = SERVER_CONFIG["endpoints"]["fit"]["path"]
+    predict_path = SERVER_CONFIG["endpoints"]["predict"]["path"]
     httpx_client = httpx.Client(
         base_url=base_url,
         timeout=TABPFN_CLIENT_TIMEOUT,
         headers={"Prior-Client-Version": get_client_version()},
-        transport=SelectiveHTTP2Transport(http2_paths=[fit_path]),
+        transport=SelectiveHTTP2Transport(http2_paths=[fit_path, predict_path]),
```
ggprior, line 224: What do you think about dropping `SelectiveHTTP2Transport` entirely? As we're only passing around metadata, this might be fine.
HTTP/2 is giving me some grief on `fit_path` as sometimes the load balancer still seems to close long running connections - will try to come up with a reproducer to confirm. IIUC HTTP/1.1 would pass the keepalive bytes we currently send to hold the connection open through without buffering.

src/tabpfn_client/estimator.py
```diff
@@ -288,24 +300,22 @@ def fit_task() -> UUID:
...
         else:
             raise NotImplementedError(
                 "Only server mode is supported at the moment for init(use_server=False)"
             )
         return self
 
-    def predict(
-        self,
-        X,
-        client_options: ClientOptions | None = None,
```
ggprior, line 307: technically a breaking change, but probably fine?

src/tabpfn_client/client.py
```diff
@@ -292,6 +293,8 @@ def fit(
             `paper_version`.
         description: str, optional
             Description of the dataset and task for the server.
+        force_refit: bool, optional
+            Whether to force refit the model even if the model has already been fitted.
         client_options : ClientOptions, optional
             Per-request options (e.g. timeout, headers) for the fitting API call
```
ggprior: is it accurate? seems passed through to the predict call as well

ggprior (approved):
LGTM, question around selective transport.

src/tabpfn_client/estimator.py
simo-prior: Yes it only breaks the beta versions.

PR #268, opened 2026-05-04
chore: align client interface to OSS interface

ggprior:
* Allow `model_path=auto` (how the option is called in OSS package), still retain `default` for backwards compat
* Set `random_state=0` to match OSS package (note this is technically a breaking change)
* Set `ignore_pretraining_limits=False` (up for discussion if we actually want to keep this as-is though, curious to your thoughts @brendan-priorlabs  -- as we handle limits already in the server.

https://linear.app/priorlabs/issue/ENG-663/snapshot-align-client-and-oss-package-interface

### DIFF
--- a/src/tabpfn_client/estimator.py
+++ b/src/tabpfn_client/estimator.py
@@ -170,7 +170,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator, TabPFNModelSelection):
         softmax_temperature: float = 0.9,
         balance_probabilities: bool = False,
         average_before_softmax: bool = False,
-        ignore_pretraining_limits: bool = False,
+        ignore_pretraining_limits: bool = True,
         inference_precision: Literal["autocast", "auto"] = "auto",
         random_state: Optional[
             Union[int, np.random.RandomState, np.random.Generator]
@@ -217,11 +217,13 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator, TabPFNModelSelection):
              predictive performance when there are many classes or when calibrating the
              model's confidence. This is only applied when predicting during a
              post-processing.
-        ignore_pretraining_limits: bool, default=False
+        ignore_pretraining_limits: bool, default=True
             Whether to ignore the pre-training limits of the model. The TabPFN models
             have been pre-trained on a specific range of input data. If the input data
             is outside of this range, the model may not perform well. You may ignore
             our limits to use the model on data outside the pre-training range.
+            Defaults to True (vs False in the OSS package): the server enforces its
+            own capacity limits, so the OSS check is redundant and stricter.
         inference_precision: "autocast" or "auto", default="auto"
             The precision to use for inference. This can dramatically affect the
             speed and reproducibility of the inference.
--- a/src/tabpfn_client/estimator.py
+++ b/src/tabpfn_client/estimator.py
@@ -49,6 +49,11 @@ DEFAULT_V2_MODEL_PATH = "v2_default"
 DEFAULT_V2_5_MODEL_PATH = "v2.5_default"
 DEFAULT_V3_MODEL_PATH = "v3_default"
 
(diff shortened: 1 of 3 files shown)


src/tabpfn_client/estimator.py
```diff
@@ -147,7 +155,7 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator, TabPFNModelSelection):
         "v2.5_real",
         "v2.5_variant",
         DEFAULT_V2_MODEL_PATH,
-        "default",
```
brendan-priorlabs, line 150: I don't think we want this change. It's essentially our public list of permissible models and we're still handling it. Keep both? btw, some of the extensions like `tabpfn-time-series` do use this

src/tabpfn_client/estimator.py
```diff
@@ -206,16 +217,17 @@ def __init__(
              predictive performance when there are many classes or when calibrating the
              model's confidence. This is only applied when predicting during a
              post-processing.
-        ignore_pretraining_limits: bool, default=True
+        ignore_pretraining_limits: bool, default=False
```
brendan-priorlabs: This is probably more correct, but if the main package hasn't set the limits correctly, we'll get failures. We can try and see what happens. Guessing there's a decent chance we'll need to revert. Maybe leave a `TODO: Verify that we don't have any artificially low limits`?

src/tabpfn_client/estimator.py
brendan-priorlabs, line 187: I see a couple more that might be possible to add. https://github.com/PriorLabs/TabPFN/blame/e22bdb4315b5abc17ae0f788c268966226db2a52/src/tabpfn/classifier.py#L231

It's confusing though, there are some on there that just don't apply, so use best judgement of course

brendan-priorlabs (approved):
Approving now, but please see that `default` comment!

src/tabpfn_client/estimator.py
ggprior: ack'd, keeping it in this list

src/tabpfn_client/estimator.py
ggprior: Yeah, on those, debated whether to add anything else here but all are not great fits, except possibly the first one

- `categorical_features_indices` -- used to be there but removed in #62 . Likely due to categorical logic in API (?)
- `eval_metric` + `tuning_config` - could be implemented client-side (this is normal optimization using a hold-out split). However we already have the evaluation metric for enhanced fit mode -- could it be confusing?
- `save_fit_state` / `load_fitted_tabpfn_model` -- not applicable

I'm guessing the deeper question is whether we want to continue best-effort like this, or enforce 100% compatible interface (accepting that some options like `eval_metric` will become no-ops in client (we could warn for no-op configs, would buy us that code is copy-pasteable and likely would work when changing `import tabpfn` to `import tabpfn_client`

brendan-priorlabs: Thanks for checking! Can't be 100%. There will for sure be server-incompatible options. Can address this in a smart way in the long-term with two (three?) different interfaces that we implement as appropriate depending on package.

For `eval_metric` specifically I see that the name for the enhanced metric is already distinct, so we're in good shape for adding that when ready. +1 to skipping for now

ggprior: note: reverting `ignore_pretraining_limits` -- OSS code would reject v3 traffic for large data `tabpfn/inference_config.py`: Model v2 `MAX_SAMPLES: 10,000, MAX_FEATURES: 500, MAX_CLASSES: 10`, 2.5/2.6 `50000/2000/10`.
We want to keep our overrides for v3 limits.
Could be aligned once `tabpfn/validation.py` is updated.

src/tabpfn_client/estimator.py
ggprior: reverted; see below for rationale

PR #271, opened 2026-05-10
Add 2.6 model

brendan-priorlabs:


ggprior (approved):
nice to have for completeness, it's a win as we ship 2.6 weights in the server works anyway

PR #286, opened 2026-05-29
Add SageMaker BYOC client (tabpfn_client.sagemaker)

ggprior:
Adds a new `tabpfn_client.sagemaker` submodule providing scikit-learn-style `TabPFNClassifier` and `TabPFNRegressor` that proxy through `boto3.client("sagemaker-runtime").invoke_endpoint` to a TabPFN SageMaker BYOC endpoint

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,42 @@ and login (on another machine) using your access token, skipping the interactive
 tabpfn_client.set_access_token(token)
 ```
 
+## AWS SageMaker (BYOC)
+
+If you've subscribed to the TabPFN AWS Marketplace listing and deployed the container to a SageMaker real-time endpoint, you can invoke it through `tabpfn_client.sagemaker` using a near-identical scikit-learn surface. There is no PriorLabs API token in this path — you authenticate to your own AWS account, and `predict` calls are billed by AWS SageMaker rather than against your TabPFN usage allowance.
+
+Install with the optional `sagemaker` extra to pull in `boto3`:
+
+```bash
+pip install --upgrade 'tabpfn-client[sagemaker]'
+```
+
+Then point the estimator at your endpoint:
+
+```python
+from tabpfn_client.sagemaker import TabPFNClassifier, TabPFNRegressor
+from sklearn.datasets import load_breast_cancer
+from sklearn.model_selection import train_test_split
+
+X, y = load_breast_cancer(return_X_y=True)
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.5, random_state=42)
+
+clf = TabPFNClassifier(
+    endpoint_name="your-sagemaker-endpoint-name",
+    region_name="us-east-1",
+)
+clf.fit(X_train, y_train)
+clf.predict(X_test)
+clf.predict_proba(X_test)
+```
+
+Notes:
+
+- AWS credentials are resolved through the standard `boto3` credential chain (env vars, `~/.aws/credentials`, instance profile, SSO, etc.). Pass `boto_session=session` to use an explicit `boto3.Session`.
+- `fit()` is local: TabPFN is in-context, so the estimator just keeps `X_train` / `y_train` and ships them on each `predict*` call. There is no separate training job.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -114,6 +114,55 @@ and login (on another machine) using your access token, skipping the interactive
...
+clf.fit(X_train, y_train)
+clf.predict(X_test)
+clf.predict_proba(X_test)
+```
+
+Notes:
+
+- AWS credentials are resolved through the standard `boto3` credential chain (env vars, `~/.aws/credentials`, instance profile, SSO, etc.). Pass `boto_session=session` to use an explicit `boto3.Session`.
+- `fit()` is local: TabPFN is in-context, so the estimator just keeps `X_train` / `y_train` and ships them on each `predict*` call. There is no separate training job.
+- Set `use_kv_cache=True` to opt into the v3 KV-cache path on the server: the first round-trip uploads training data and captures a `model_id`; subsequent `predict*` calls reference that id and skip the training upload. This trades a stateful endpoint for lower per-call latency and payload size.
```
brendan-priorlabs: > This trades a stateful endpoint for lower per-call latency and payload size.

I think this statement will be confusing for users. When should they avoid setting setting the KV cache might be more helpful. (nvm if `stateful_endpoint` is literally Sagemaker config or something!)

src/tabpfn_client/sagemaker/estimator.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def predict_proba(self, X: Any) -> np.ndarray:
+        result = self._invoke(X, output_type="probas")
+        return np.asarray(result["prediction"])
+
+
+class TabPFNRegressor(_SagemakerBase, RegressorMixin):
+    """TabPFN regressor backed by a SageMaker real-time endpoint.
+
+    `output_type` defaults to "mean"; pass "median", "mode", "full", or
```
brendan-priorlabs: nit: Method comment? (Had an LLM do this to me over the weekend and it was driving me nuts)

src/tabpfn_client/sagemaker/estimator.py
```diff
@@ -0,0 +1,342 @@
+#  Copyright (c) Prior Labs GmbH 2026.
+#  Licensed under the Apache License, Version 2.0
+"""scikit-learn estimators for the TabPFN SageMaker BYOC endpoint.
+
+Mirrors the `tabpfn_client.TabPFNClassifier` / `TabPFNRegressor` surface;
+each `predict*` call dispatches via `boto3.client("sagemaker-runtime")`
+against your SageMaker endpoint. `fit()` is local (TabPFN is in-context).
```
brendan-priorlabs: > `fit()` is local (TabPFN is in-context).

I think this is misleading. `fit` is certainly not local from the perspective of the client, right?

brendan-priorlabs (approved):
Good stuff! Very nice usability improvement. Left just a couple minor comments

README.md
ggprior: yeah, the README is not great. I'll human-write this.

src/tabpfn_client/sagemaker/estimator.py
ggprior: `fit` is local from the perspective of the client. It only is there to satisfy the sklearn interface, and keep track of our model_id housekeeping for the KV cache.
This is because Sagemaker only gives us a single endpoint to work with, `/invocations`. 
And actually part of what we do with `.fit()` in the server is handled by the sagemaker layer already (upload to s3 bucket, store results to s3 bucket, etc)

brendan-priorlabs: Ah, so even the thinking fit happens in (the first) predict? I'd still nitpick that "TabPFN is in-context" is confusing because, while true, `fit` is not local in the API, so it's not the pertinent fact here

src/tabpfn_client/sagemaker/estimator.py
ggprior: nit response: though weird it's not unheard of in python to put `__init__()` args into the class-level docstring, because that's typically what IDEs show on autocomplete and `help(Class)`

brendan-priorlabs: `__init__()` args of course, but `output_type` is an arg only to `predict`!

PR #288, opened 2026-06-01
Use server-generated api models

simo-prior:


brendan-priorlabs (commented):
Any chance we can instead put `models_gen.py` into `api_models.py` so we don't break anyone importing them. No need to leak the `gapi` name either?

simo-prior:
> Any chance we can instead put `models_gen.py` into `api_models.py` so we don't break anyone importing them. No need to leak the `gapi` name either?

Can do that, need to publish the package first.

Edit: addressed.

src/tabpfn_client/api_models.py
```diff
@@ -1,148 +1,312 @@
...
-class PredictResponse(BaseModel):
-    prediction: Prediction
-    metadata: Metadata
+class PrepareTrainSetUploadResponse(BaseModel):
+    train_set_upload_id: UUID
+    x_train_info: FileUploadInfo
+    y_train_info: FileUploadInfo
+
+
+class DuplicateTrainSetErrorResponse(BaseModel):
```
brendan-priorlabs, line 317: Previously this subclassed `ErrorResponse`. Any issues with losing that?

src/tabpfn_client/estimator.py
```diff
@@ -764,6 +805,30 @@ def predict_task() -> PredictionResult:
 
         return output
 
+    def _get_tabpfn_config(self) -> RegressorTabPFNConfig:
+        init_params = self.get_params()
+        # Drop params left at their `None` default so the config model applies its
+        # own field defaults (e.g. ignore_pretraining_limits) rather than failing
+        # validation on a `None` for a non-optional field.
```
brendan-priorlabs: There's an inconsistency here between the regressor and classifier w.r.t. dropping the `None`s. They should behave the same unless there is a very strong reason

src/tabpfn_client/estimator.py
```diff
@@ -699,16 +735,19 @@ def predict(
         array-like or dict
             The predicted values.
         """
+        kwargs = locals()
```
brendan-priorlabs: Might be good to add a little warning comment to keep this at the top of the function, so that things don't break in a refactor

src/tabpfn_client/api_models.py
```diff
@@ -1,148 +1,312 @@
...
+    softmax_temperature: float | None = None
+    average_before_softmax: bool | None = None
+    random_state: int | None = None
+    inference_config: dict[str, Any] | None = Field(
+        default=None, description="Refer to tabpfn.inference_config.InferenceConfig for more details."
+    )
+    inference_precision: (
+        Annotated[Literal["autocast", "auto"] | str, Field(union_mode="left_to_right")] | None
+    ) = None
+    ignore_pretraining_limits: bool = True
```
brendan-priorlabs: Why `True` here instead of `None`? (Separately, this is a behavior change for the regressor specifically. Probably good that we're unifying behavior...)

src/tabpfn_client/estimator.py
```diff
@@ -85,12 +96,12 @@ def _validate_model_name(cls, model_name: str) -> None:
...
     def _model_name_to_path(
-        cls, task: Literal["classification", "regression"], model_name: str
-    ) -> Optional[str]:
+        cls, task: Literal["classification", "regression"], model_name: str | None
+    ) -> str | None:
         cls._validate_model_name(model_name)
         model_name_task = "classifier" if task == "classification" else "regressor"
         # Let the server pick the default model when the caller defers to us.
-        if model_name in _AUTO_MODEL_PATH_ALIASES:
+        if model_name is None or model_name in _AUTO_MODEL_PATH_ALIASES:
```
brendan-priorlabs: I think there is another place where this check comes up. `_is_thinking_supported_model_path` Can we just add `None` to that list to remove all this duplicate code?

src/tabpfn_client/estimator.py
```diff
@@ -178,22 +167,23 @@ class TabPFNClassifier(ClassifierMixin, BaseEstimator, TabPFNModelSelection):
...
+        # start: tabpfn_config
+        model_path: str | None = None,
+        categorical_features_indices: list[int] | None = None,
+        n_estimators: int | None = None,
+        softmax_temperature: float | None = None,
+        balance_probabilities: bool | None = None,
+        average_before_softmax: bool | None = None,
+        ignore_pretraining_limits: bool | None = None,
+        inference_precision: Literal["autocast", "auto"] | None = None,
+        random_state: int | None = None,
```
brendan-priorlabs: Could you double check this one specifically since it could surprise users rely on reproducibility?

brendan-priorlabs (approved):
Alright, looks good overall! For some final manual testing could you be sure to:

1. Run a regular fit + predict
2. Run with text
3. Run with thinking (I think this would have caught the bug I mention below)

There are lots of config options otherwise of course, but covering at least those will get us pretty far

src/tabpfn_client/api_models.py
simo-prior: Yes the codegen tool removes inheritance and duplicates fields instead. I guess it's easier to write the generation logic this way. 

For the moment I don't see an issue with that, if we find a need in the future (eg. matching on all subclasses of ErrorResponse) we can adjust.

src/tabpfn_client/estimator.py
simo-prior: Cannot spot the inconsistency. That said the comment is too verbose here: `Nones are treated as unset` is enough.

src/tabpfn_client/estimator.py
simo-prior: Addressed.

src/tabpfn_client/api_models.py
simo-prior: Good point, `ignore_pretraining_limits` is *not* pass-through on the server, meaning the server overrides tabpfn default. That said on the client it should be pass-through wrt to the server. Addressing.

src/tabpfn_client/estimator.py
simo-prior: Addressed.

src/tabpfn_client/estimator.py
simo-prior: Double checked, the default values will remain the same as `tabpfn` defaults are the same we had on the client. The only exception is `ignore_pretraining_limits` which defaulted to `True` in the client and defaults to `False` in `tabpfn`. But on the server we override it to `True` so the end result is the same.

PR #292, opened 2026-06-04
Setup trunk w/ pyright

simo-prior:


src/tabpfn_client/client.py
```diff
@@ -113,13 +113,18 @@ class RetryableServerError(Exception):
 
 
 class SensitiveDataFilter(logging.Filter):
-    def filter(self, record):
-        if "password" in record.getMessage():
-            original_query = str(record.args[1])
+    def filter(self, record: logging.LogRecord) -> bool:
```
simo-prior, line 116: **Behavior change (safe hardening):** redaction now requires `args` to be a tuple of length ≥ 2. Previously a log record whose message contained `password` but whose `args` wasn't a ≥2-tuple would raise `IndexError`/`TypeError` *inside the log filter*. Now it cleanly skips; the normal redaction path is unchanged.

src/tabpfn_client/config.py
```diff
@@ -127,7 +127,10 @@ def get_access_token() -> str:
     :return: The access token string used for API requests.
     """
     init()
-    return ServiceClient.get_access_token()
+    access_token = ServiceClient.get_access_token()
+    if access_token is None:
```
simo-prior, line 131: **Behavior change (safe):** `get_access_token()` now raises `CONNECTION_ERROR` instead of returning `None`. Callers expect a usable token; returning `None` just deferred the failure downstream. Fail-fast is the right contract, and the return type (`str`) is now honored.

src/tabpfn_client/service_wrapper.py
```diff
@@ -88,17 +88,19 @@ def set_token_by_registration(
...
     @classmethod
-    def set_token_by_login(cls, email: str, password: str) -> tuple[bool, str]:
+    def set_token_by_login(
+        cls, email: str, password: str
+    ) -> tuple[str | None, str, int]:
         access_token, message, status_code, _ = ServiceClient.login(email, password)
 
         if access_token is None:
-            return False, message, status_code
+            return None, message, status_code
```
simo-prior, line 97: **Return-contract change:** on login failure the first tuple element is now `None` instead of `False` (type is now `tuple[str | None, str, int]`). The sole caller (`prompt_agent`) gates on `status_code == 200 and access_token is not None`, so this is behavior-preserving there — but any caller doing `is False` would break.

src/tabpfn_client/service_wrapper.py
```diff
@@ -117,6 +119,8 @@ def try_reuse_existing_token(cls) -> tuple[bool, str or None]:
             return False, access_token
 
         logger.debug(f"Reusing existing access token? {is_valid}")
+        if access_token is None:
```
simo-prior, line 122: **Behavior change (safe):** guards against `set_token(None)`. Previously, if the token validated as not-outdated but was `None`, this stored a `None` token and returned `(True, None)`; now it returns `(False, None)` and caches nothing.

src/tabpfn_client/estimator.py
```diff
@@ -405,9 +405,12 @@ def _predict(
         )
 
         validate_test_set(X, output_type, tabpfn_config.model_path)
-        X = _clean_text_features(X)
+        X_clean = _clean_text_features(X)
 
-        if "sentry-trace" not in self.client_options.headers:
+        if (
+            "sentry-trace" not in self.client_options.headers
+            and self.last_trace_id is not None
```
simo-prior, line 412: **Behavior change (safe):** the `sentry-trace` header is only set when `last_trace_id is not None`, avoiding writing a literal `None` header value. (Same change in the regressor `predict` below.)

src/tabpfn_client/estimator.py
```diff
@@ -747,9 +750,12 @@ def predict(
...
```
simo-prior, line 757: **Behavior change (safe):** same `sentry-trace` `None`-guard as the classifier — don't set the header to `None`.

src/tabpfn_client/estimator.py
```diff
@@ -829,16 +835,17 @@ def _get_predict_params(self, kwargs: dict[str, Any]) -> RegressorPredictParams:
         }
         return RegressorPredictParams.model_validate(params)
 
-    def _validate_targets(self, y) -> np.ndarray:
+    def _validate_targets(self, y) -> None:
         y_ = column_or_1d(y, warn=True)
         if sum(pd.isnull(y_)) > 0:
             raise ValueError("Input y contains NaN.")
```
simo-prior, line 843: **Behavior change (bug fix):** `model_path is None` is now treated as “defer to server”. Since `model_path` defaults to `None`, the old `V_3_IDENTIFIER in model_path` on the fallthrough would raise `TypeError: argument of type 'NoneType' is not iterable` when thinking was enabled with the default model. This makes the `None` case return `True` instead of crashing.

src/tabpfn_client/prompt_agent.py
```diff
@@ -627,7 +629,10 @@ def _choice_with_retries(cls, prompt: str, choices: list) -> str:
         return choice.lower()
 
     @classmethod
-    def _verify_user_email(cls, access_token: str) -> bool:
+    def _verify_user_email(cls, access_token: str | None) -> bool:
+        if access_token is None:
```
simo-prior, line 633: **Behavior change (safe):** `_verify_user_email` now bails out with a clear `fail(...)` + `return False` when `access_token is None`, instead of proceeding into verification with a `None` token.

simo-prior (commented):
Flagging the spots where this PR changes actual runtime behavior (vs. pure type annotations), with reasoning. All are safe-by-design — mostly None-guards that prevent latent crashes / bad state, plus one dead-branch removal. Nothing here changes a successful happy-path result.

.github/workflows/cache_trunk.yaml
```diff
@@ -0,0 +1,23 @@
...
+permissions: read-all
+
+jobs:
+  cache_trunk:
+    name: Cache Trunk
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
brendan-priorlabs: Should we start pinning these? I've read a few places that this is recommended to prevent supply chain attacks?

src/tabpfn_client/estimator.py
```diff
@@ -42,11 +42,11 @@
...
 try:
-    from torch import Tensor
-
-    TORCH_AVAILABLE = True
+    from torch import Tensor  # type: ignore
 except ImportError:
-    TORCH_AVAILABLE = False
+    Tensor = None
+
+TORCH_AVAILABLE = Tensor is not None
```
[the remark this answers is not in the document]
simo-prior: ✅ Fixed in 3920d0b. Defined a dummy `Tensor` class on `ImportError` so `Tensor` is always a type, which is exactly what `isinstance` needs. `TORCH_AVAILABLE` is now set directly in each branch.

src/tabpfn_client/estimator.py
```diff
@@ -961,41 +978,42 @@ def _clean_text_features(X):
...
-        if X.is_cuda:
-            X = X.cpu()
-
-        X = X.numpy()
-
-    if isinstance(X, np.ndarray):
-        if np.issubdtype(X.dtype, np.number):
-            return X
+    data = X
+    if Tensor is not None and isinstance(data, Tensor):
```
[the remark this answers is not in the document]
simo-prior: ✅ Fixed in 3920d0b. With `Tensor` always a type, dropped the `Tensor is not None` guard — now just `isinstance(data, Tensor)`.

src/tabpfn_client/estimator.py
```diff
@@ -405,9 +405,12 @@ def _predict(
...
         validate_test_set(X, output_type, tabpfn_config.model_path)
-        X = _clean_text_features(X)
+        X_clean = _clean_text_features(X)
 
-        if "sentry-trace" not in self.client_options.headers:
+        if (
+            "sentry-trace" not in self.client_options.headers
+            and self.last_trace_id is not None
+        ):
             self.client_options.headers["sentry-trace"] = self.last_trace_id
```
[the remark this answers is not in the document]
simo-prior: ❌ Not applied. `last_trace_id` is unconditionally initialized in `__init__` (line 296), so the `AttributeError` can't occur here. Adding `getattr` would only add defensive clutter that obscures the invariant. The unpickle-from-older-version case is speculative.

src/tabpfn_client/estimator.py
```diff
@@ -747,9 +750,12 @@ def predict(
...
```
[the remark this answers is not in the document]
simo-prior: ❌ Not applied — same reasoning as the classifier: `last_trace_id` is always set in `__init__` (line 636).

tests/unit/test_tabpfn_classifier.py
```diff
@@ -487,7 +490,7 @@ def test_only_allowed_parameters_passed_to_config(self):
 
         # Skip fitting
         classifier.fitted_ = True
-        classifier.last_fitted_train_set_id = "dummy_uid"
+        classifier.last_fitted_train_set_id = cast(UUID, "dummy_uid")
```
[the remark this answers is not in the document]
simo-prior: ✅ Fixed in 3920d0b. Replaced `cast(UUID, "dummy_uid")` with a real `UUID("00000000-0000-0000-0000-000000000000")`, removing the type lie.

tests/unit/test_tabpfn_regressor.py
```diff
@@ -481,7 +484,7 @@ def test_only_allowed_parameters_passed_to_config(self):
 
         # Skip fitting
         regressor.fitted_ = True
-        regressor.last_fitted_train_set_id = "dummy_uid"
+        regressor.last_fitted_train_set_id = cast(UUID, "dummy_uid")
```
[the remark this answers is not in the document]
simo-prior: ✅ Fixed in 3920d0b. Replaced with a real `UUID(...)`.

tests/unit/test_tabpfn_regressor.py
```diff
@@ -549,7 +552,7 @@ def test_predict_params_output_type(self):
     def test_predict_full_adds_criterion_with_optional_dependencies(self):
         regressor = TabPFNRegressor()
         regressor.fitted_ = True
-        regressor.last_fitted_train_set_id = "dummy_uid"
+        regressor.last_fitted_train_set_id = cast(UUID, "dummy_uid")
```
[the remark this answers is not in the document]
simo-prior: ✅ Fixed in 3920d0b. Replaced with a real `UUID(...)`.

tests/unit/test_tabpfn_regressor.py
```diff
@@ -589,7 +599,7 @@ def __init__(self, borders):
     def test_predict_full_missing_optional_dependencies_logs_warning(self):
         regressor = TabPFNRegressor()
         regressor.fitted_ = True
-        regressor.last_fitted_train_set_id = "dummy_uid"
+        regressor.last_fitted_train_set_id = cast(UUID, "dummy_uid")
```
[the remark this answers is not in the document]
simo-prior: ✅ Fixed in 3920d0b. Replaced with a real `UUID(...)`.

.trunk/scripts/pyright_to_sarif.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+#
+# Adapted from trunk-io/plugins:
+#   https://github.com/trunk-io/plugins/blob/main/linters/pyright/pyright_to_sarif.py
+#
+# Copyright (c) Trunk Technologies, Inc.
+# SPDX-License-Identifier: MIT
+#
+# Local modifications: wrapped the per-result parse in try/except KeyError for
```
brendan-priorlabs, line 9: What is this file for? This sort of LLM description is not helpful

README.md
```diff
@@ -266,10 +266,10 @@ This project is licensed under the Apache License 2.0 — see the [LICENSE](LICE
 <details>
 <summary><b>Setup, build, and release instructions</b></summary>
 
-To encourage better coding practices, `ruff` has been added to the pre-commit hooks. This will ensure that the code is formatted properly before being committed. To enable pre-commit (if you haven't), run the following command:
+To encourage better coding practices, linting and formatting are managed with [Trunk](https://trunk.io) (running `ruff` and `basedpyright`). To check your changes, run:
 
 ```bash
-pre-commit install
+trunk check
```
brendan-priorlabs, line 272: Can you give full instructions including install and format?

README.md
```diff
@@ -266,10 +266,10 @@ This project is licensed under the Apache License 2.0 — see the [LICENSE](LICE
...
 ```
 
 Additionally, it is recommended that developers install the ruff extension in their preferred editor. For installation instructions, refer to the [Ruff Integrations Documentation](https://docs.astral.sh/ruff/integrations/).
```
brendan-priorlabs: Still true? How do we get the versions to match?

src/tabpfn_client/config.py
brendan-priorlabs: Is this actually true? Didn't we have some weird exception for dev users or in the smoke test?

src/tabpfn_client/estimator.py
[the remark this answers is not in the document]
simo-prior: ⚠️ Correction: reverted in ee267f1. The dummy `Tensor` class actually **regresses basedpyright** — when torch is installed, basedpyright unions the dummy with the real `torch.Tensor`, and the attribute-less dummy wins for attribute access, producing 5 `reportAttributeAccessIssue` errors (`requires_grad`/`detach`/`is_cuda`/`cpu`/`numpy`). Restored the original `Tensor = None` form, which passes `trunk check` cleanly.

src/tabpfn_client/estimator.py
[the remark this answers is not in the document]
simo-prior: ⚠️ Correction: reverted in ee267f1 along with the dummy-class change — restored the `Tensor is not None and isinstance(...)` guard. See the note on the import block: the dummy class introduced 5 new basedpyright errors, so the original form is the one that type-checks cleanly.

src/tabpfn_client/prompt_agent.py
```diff
@@ -627,7 +629,10 @@ def _choice_with_retries(cls, prompt: str, choices: list) -> str:
...
+            fail("No access token available for email verification.")
```
brendan-priorlabs, line 634: Why would we have an access token at this state?

brendan-priorlabs (commented):
Could you comment a bit on the testing plan for this? I'm happy about the change, but a bit worried since so many places were touched, e.g., not allowing `None` tokens might be a bug

.trunk/scripts/pyright_to_sarif.py
simo-prior: I copied it over from `tabpfn-server`, we already use it there. Not LLM-generated, it's taken from trunk repo to parse pyright output in a trunk-compatible way.

.github/workflows/cache_trunk.yaml
simo-prior: Addressed.

README.md
simo-prior: Hmm let's remove it, the link doesn't point to IDE's integrations and ofc people can set their editor as they wish.

(if you refer to ruff itself, it is installed in the venv)

README.md
simo-prior: I changed the link above to the actual installation page, `trunk check` does both type checking and formatting.

src/tabpfn_client/config.py
simo-prior: Not in the smoke test at least, indeed returning None just postpones the exception making it less descriptive of the actual problem.

src/tabpfn_client/prompt_agent.py
simo-prior: The access token can be already minted as the user does exist in the db, it's just flagged.

.github/workflows/cache_trunk.yaml
```diff
@@ -0,0 +1,23 @@
...
+permissions: read-all
+
+jobs:
+  cache_trunk:
+    name: Cache Trunk
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6.0.3
```
brendan-priorlabs: Somewhat absurdly, tags aren't enough. The pins need to be full SHAs. See, e.g., https://news.ycombinator.com/item?id=47475888 We probably also need to think about how to upgrade, but that can come later of course

brendan-priorlabs (approved):
Thanks for the followups! Left one clarification worth addressing

.github/workflows/cache_trunk.yaml
simo-prior: Addressed.

Issue #285, opened 2026-05-27
Relax upper bound on `scikit-learn` dependency


Hunterlige:
Hi

I noticed the project currently pins scikit-learn to version 1.6.1 as an upper limit.

Since scikit-learn 1.6.1 (https://github.com/scikit-learn/scikit-learn/releases/tag/1.6.1) was released on Jan 10, 2025, I wanted to ask whether there is a specific compatibility concern that requires keeping this restriction in place.

Would it be possible to relax or widen the version constraint to allow newer scikit-learn releases? This would help downstream users avoid dependency conflicts in environments that already use more recent versions.

If there are known incompatibilities with newer versions, documenting them would also be very helpful.

simo-prior:
Thank you for the suggestion @Hunterlige, we just upgraded :)

PR #305, opened 2026-06-11
Relax scikit-learn upper bound to 1.9.0, drop Python 3.9

simo-prior:
## Description

The `scikit-learn` cap has been stuck at 1.6.1 (Jan 2025) — see #285. There was no actual incompatibility: the client only uses stable sklearn API (`BaseEstimator`, mixins, `check_is_fitted`, `column_or_1d`, `check_classification_targets`). The cap went stale because dependabot cannot propose versions whose Python requirement is narrower than the project's floor: sklearn 1.7 requires Python ≥3.10 and 1.8+ requires ≥3.11, while we declared ≥3.9.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,6 @@ updates:
           - "ruff"
       test-dependencies:
         patterns:
-          - "scikit-learn"
           - "respx"
 
   # GitHub Actions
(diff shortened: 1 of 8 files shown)


safaricd (approved):
LGTM

PR #316, opened 2026-06-18
Support async fit calls

simo-prior:
- [x] Support for async fit calls
- [x] Determine which mode to use depending on train set size
- [x] Env var to override default behavior

Note: `force_fit` is removed (still supported on the server for backward compatibility, but it will become a no-op). It is no longer needed as every `/fit` call now generates a different `fitted_train_set_id` rather than trying to identify different calls under the same id: it's no longer necessary to provide an option to override that behavior.


src/tabpfn_client/api_models.py
```diff
@@ -41,42 +41,11 @@ def __get_pydantic_core_schema__(cls, _source, _handler):
         return core_schema.no_info_after_validator_function(cls, core_schema.str_schema())
 
 
-class PredictionTask(str, Enum):
```
simo-prior, line 44: Unfortunately the diff in this file is mostly related to changes in the ordering algo on the SDK generator. Hopefully future diffs will only point to actual additions/deletions.

src/tabpfn_client/client.py
```diff
@@ -334,10 +335,17 @@ def fit(
...
+
         limits = cls.get_model_limits()
         if limits is not None:
             for name, df in [("x_train", df_X), ("y_train", df_y)]:
                 mem_usage = df.memory_usage(deep=True).sum()
+                if (
+                    mem_usage <= _ASYNC_MODE_ENABLED_ABOVE_SIZE_MB
+                    and not force_async_enabled()
+                ):
+                    async_mode = False
```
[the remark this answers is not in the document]
simo-prior: Addressed

src/tabpfn_client/client.py
```diff
@@ -436,8 +444,75 @@ def fit(
...
+    @classmethod
+    def _get_fit_status(
+        cls,
+        fitted_train_set_id: UUID,
+        headers: dict[str, str] | None = None,
+    ) -> httpx.Response:
+        return cls.httpx_client.get(
+            url=f"/tabpfn/fit/{fitted_train_set_id}",
+            headers=headers,
+        )
```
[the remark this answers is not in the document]
simo-prior: Addressed

brendan-priorlabs:
It looks like this PR also removes the force refit option. Can you please document the implications of that in the PR description? (I know you and I talked about that, but my memory of the details is hazy and everyone else will need context too.)

simo-prior:
> It looks like this PR also removes the force refit option. Can you please document the implications of that in the PR description? (I know you and I talked about that, but my memory of the details is hazy and everyone else will need context too.)

@brendan-priorlabs addressed

src/tabpfn_client/api_models.py
```diff
@@ -200,12 +222,23 @@ class FitRequest(BaseModel):
...
+        description="Submit the fit and return immediately with status=pending; poll GET /tabpfn/fit/{fitted_train_set_id} for the outcome. Defaults to the blocking behaviour.",
     )
 
 
 class FitResponse(BaseModel):
     fitted_train_set_id: UUID
+    status: Annotated[FitStatus | UnknownEnum, Field(union_mode="left_to_right")]
+
+
+class FitStatusResponse(BaseModel):
```
safaricd, line 238: Curious - why do we need the `FitStatusResponse` and `FitResponse`, if the only diff is `error` which is nullable anyway? Couldn't we wrap this under a single type?

src/tabpfn_client/api_models.py
```diff
@@ -217,6 +250,13 @@ class GetModelLimitsResponse(BaseModel):
     dataset_max_size_bytes: int
 
 
+class NotFoundErrorResponse(BaseModel):
```
safaricd, line 253: Is the `NotFoundErrorResponse` used whenever GAPI returns back a 404 HTTP status code? Or is there some other motive?

src/tabpfn_client/client.py
```diff
@@ -334,15 +336,24 @@ def fit(
...
                 if mem_usage > limits.dataset_max_size_bytes:
                     raise ValueError(
                         f"In-memory size of {name} ({mem_usage} bytes) exceeds "
                         f"the server limit of {limits.dataset_max_size_bytes} bytes."
                     )
+            if (
+                total_mem_usage <= _ASYNC_MODE_ENABLED_ABOVE_SIZE_BYTES
+                and not force_async_enabled()
+            ):
+                async_mode = False
```
safaricd, line 356: Could you please walk me through the decision for disabling async with based on the host's local memory usage? Memory usage should be the same regardless of whether we use async or sync mode - or am I missing something?

src/tabpfn_client/client.py
```diff
@@ -436,8 +447,85 @@ def fit(
...
+    ) -> None:
+        """Poll ``GET /tabpfn/fit/{id}`` until the fit reaches a terminal state.
+
+        Returns once the fit is COMPLETED. Raises if the fit FAILED, or
+        ``TimeoutError`` if it does not reach a terminal state within ``timeout``
+        seconds. A failed fit is reported by the server as HTTP 200 with
+        ``status=failed`` (only an unknown/foreign id is a 404), so the status
+        field — not the HTTP code — is what we inspect here.
+        """
+        deadline = time.monotonic() + timeout
```
safaricd, line 478: Instead of a hard deadline - perhaps a v2 change - the timeout should be relative to progress? For example, a fit job running for `TABPFN_CLIENT_POLL_TIMEOUT` should time out. However, a fit job running for `TABPFN_CLIENT_POLL_TIMEOUT` but making progress should be treated differently?

src/tabpfn_client/api_models.py
simo-prior: Keep in mind these models are generated from server models. On the server `*Request` and `*Response` are always unique to the endpoint, as a result here they are also distinct. One of the reasons is to make sure we don't involuntarily share fields across endpoints.

src/tabpfn_client/api_models.py
simo-prior: Not necessarily, this model is explicitly specified in the `response_models` field on the server. Otherwise you will have a generic `ErrorResponse`.

src/tabpfn_client/client.py
simo-prior: You mean `PENDING` but actually stuck somehow? We won't be able to tell, we don't have heartbeats for now, just a deadline.

That said there's definitely something with defaults (this and the memory usage topic) that needs to be fine tuned in a later PR (adding a ticket to the project).

src/tabpfn_client/client.py
simo-prior: We're looking for some heuristics to decide on async mode, the train set in-memory size could be it as it should be a good predictor of request duration.
That said not 100% convinced on this, will revisit in the future 👍

src/tabpfn_client/api_models.py
safaricd: > One of the reasons is to make sure we don't involuntarily share fields across endpoints.

I get it, and that's good reasoning - nevertheless, I still think these should/could be unified under a single `FitResponse` class, because an `error` may be returned from any fit endpoint - async or sync. 

Not super critical, but something to keep in mind for the future because it will cut down on some of the redundancy.

src/tabpfn_client/client.py
safaricd: Yeah - something in line of heartbeats, indicating that a fit job is getting processes, instead of being stuck - because right not, we treat both equally with the `PENDING` status.

Not critical, but something to keep in mind for the future.

safaricd (approved):

PR #332, opened 2026-07-07
feat(sagemaker): pass through async InvocationTimeoutSeconds

ggprior:
The SageMaker async estimators previously invoked `invoke_endpoint_async` with no `InvocationTimeoutSeconds`, so every async request rode the service default and long-running requests could be failed by the platform before completing with no way to extend the window from the client. This adds an optional `invocation_timeout_s` constructor argument that, when set, is forwarded as `InvocationTimeoutSeconds` (the server-side 1–3600s per-request cap); it defaults to `None` so behaviour is unchanged when unset, and it is deliberately distinct from `async_timeout_s`, which only bounds how long the client polls S3 for the result. A mocked unit test covers both the pass-through and the default-omitted cases and needs no real boto3.

### DIFF
--- a/src/tabpfn_client/sagemaker/estimator.py
+++ b/src/tabpfn_client/sagemaker/estimator.py
@@ -79,6 +79,7 @@ class _SagemakerBase(BaseEstimator):
         s3_prefix: str = "async-io",
         async_poll_interval_s: float = 2.0,
         async_timeout_s: float = 60 * 60,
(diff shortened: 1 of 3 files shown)


ggprior:
Superseded by #333 — same commits, branch renamed to `georg/eng-841-invocation-passthrough` to link ENG-841. (GitHub's branch rename closed this PR instead of retargeting it.)

PR #341, opened 2026-07-15
Normalize quantiles predict output to match local tabpfn

eliott-kalfon:
## Problem

`predict(output_type="quantiles")` returns different container types depending on the backend:

- **Local `tabpfn`**: a `list` of per-quantile 1D arrays, e.g. `[arr(n_samples,), arr(n_samples,), ...]`
- **`tabpfn-client`**: a single stacked 2D `np.ndarray` of shape `(n_quantiles, n_samples)`

Downstream code that consumes quantile predictions has to special-case each backend, and assuming the wrong type silently breaks indexing/shape logic.

Verified against `tabpfn==8.1.0` and `tabpfn-client==0.3.3`.

## Fix

In `TabPFNRegressor.predict`, when `output_type="quantiles"`, convert the server's stacked array into a list of per-quantile arrays so both backends agree on the return type.

## Test

Added `test_predict_quantiles_returns_list_matching_local_tabpfn` asserting the output is a `list` of `n_quantiles` arrays each of shape `(n_samples,)`.

### DIFF
--- a/tests/unit/test_tabpfn_regressor.py
+++ b/tests/unit/test_tabpfn_regressor.py
@@ -581,6 +581,26 @@ class TestTabPFNRegressorInference(unittest.TestCase):
             self.assertIsInstance(arr, np.ndarray)
             self.assertEqual(arr.shape, (20,))
 
+    def test_predict_single_quantile_returns_list_of_one_array(self):
+        """A single quantile may arrive squeezed to 1D; it must still become a
+        list of one (n_samples,) array, matching local tabpfn."""
+        regressor = TabPFNRegressor()
+        regressor.fitted_ = True
+        test_X = np.random.randn(20, 5)
+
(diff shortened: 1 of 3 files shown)


tests/unit/test_tabpfn_regressor.py
ggprior, line 1: random thought here, it could be pretty cool to `import tabpfn` in tests here and see if return types adhere to the same interface.
Not for this PR, but @simo-prior what do you think? As a quick solution of the ongoing align interface work.

ggprior (approved):
LGTM
Let's do a minor version bump here in pyproject.toml before merging

tests/unit/test_tabpfn_regressor.py
simo-prior: Let's think about it, maybe a good additional check but we need to keep in mind that the tabpfn version should be the same one we pinned on the server and we won't necessarily support the entire interface.

PR #342, opened 2026-07-18
TabPFN interface for on-prem

safaricd:
### Change Description

Implement`TabPFNClassifier` and `TabPFNRegressor` interfaces for self-hosted TabPFN.

### DIFF
--- a/src/tabpfn_client/endpoints/__init__.py
+++ b/src/tabpfn_client/endpoints/__init__.py
@@ -1,0 +1,11 @@
+#  Copyright (c) Prior Labs GmbH 2026.
+#  Licensed under the Apache License, Version 2.0
+"""HTTP client for a self-hosted TabPFN inference container.
+
+from tabpfn_client.endpoints import TabPFNClassifier, TabPFNRegressor
+"""
+
+from tabpfn_client.endpoints.estimator import TabPFNClassifier, TabPFNRegressor
+
+
+__all__ = ["TabPFNClassifier", "TabPFNRegressor"]
--- a/src/tabpfn_client/endpoints/estimator.py
+++ b/src/tabpfn_client/endpoints/estimator.py
@@ -1,0 +1,251 @@
+#  Copyright (c) Prior Labs GmbH 2026.
+#  Licensed under the Apache License, Version 2.0
+"""scikit-learn estimators for a TabPFN inference container reached over HTTP.
+
+Mirrors the `tabpfn_client.TabPFNClassifier` / `TabPFNRegressor` surface;
+each `predict*` call POSTs to the user-supplied `endpoint_url` (the full
+scoring URL, including the `/predict` path). `fit()` does not call the
+endpoint — it just stores `X` / `y` on the estimator. The training data
+is shipped to the endpoint on the next `predict*` call, where the actual
+fit runs.
+
+Auth is optional: when `api_key` is set it is sent as `Bearer <api_key>`,
+otherwise no `Authorization` header is added.
+
+Cached predict is a per-call option: pass `model_id` to any `predict*`
+call to send `context.model_id` and omit the training data, letting the
+endpoint reuse an already-fit model. `fit()` is not required in that
+case. When the endpoint returns a `model_id` in its response, it is
+exposed as `self.model_id_` so callers can pass it forward. `model_path`
+is similarly optional and only sent when set; some deployments reject
+overrides.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, List, Literal, Optional
+
+import httpx
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, ClassifierMixin, RegressorMixin
+from sklearn.utils.validation import check_is_fitted
+
+
+def _to_jsonable(X: Any) -> list:
+    """Coerce numpy / pandas inputs to plain Python lists for JSON."""
+    if isinstance(X, pd.DataFrame):
+        return X.values.tolist()
+    if isinstance(X, pd.Series):
+        return X.tolist()
+    return np.asarray(X).tolist()
+
+
+class _EndpointBase(BaseEstimator):
+    """Shared HTTP plumbing for the endpoint-backed TabPFN estimators."""
+
+    _TASK: str = ""  # overridden by subclasses
+
+    def __init__(
+        self,
+        endpoint_url: str,
+        api_key: Optional[str] = None,
+        extra_headers: Optional[Dict[str, str]] = None,
+        model_path: Optional[str] = None,
+        n_estimators: int = 8,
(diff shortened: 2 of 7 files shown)


src/tabpfn_client/hosted/estimator.py
```diff
@@ -0,0 +1,251 @@
...
+            return client
+        client = httpx.Client(timeout=self.timeout_s)
+        self._cached_client = client
+        return client
+
+    def __getstate__(self) -> Dict[str, Any]:
+        # httpx.Client isn't pickleable; strip the cache for sklearn pickling.
+        state = self.__dict__.copy()
+        state.pop("_cached_client", None)
+        return state
```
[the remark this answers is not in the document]
safaricd: Gemini is wrong - `copy.deepcopy` respects `__getstate__` when no `__deepcopy__` is defined, and the existing `__getstate__` already strips `_cached_client`.

src/tabpfn_client/hosted/estimator.py
```diff
@@ -0,0 +1,251 @@
...
+from sklearn.utils.validation import check_is_fitted
+
+
+def _to_jsonable(X: Any) -> list:
+    """Coerce numpy / pandas inputs to plain Python lists for JSON."""
+    if isinstance(X, pd.DataFrame):
+        return X.values.tolist()
+    if isinstance(X, pd.Series):
+        return X.tolist()
+    return np.asarray(X).tolist()
```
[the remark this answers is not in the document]
safaricd: Sure, good minor suggestion.

src/tabpfn_client/endpoints/estimator.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+
+    _TASK = "regression"
+
+    def predict(
+        self,
+        X: Any,
+        output_type: str = "mean",
+        quantiles: Optional[list] = None,
+        model_id: Optional[str] = None,
```
simo-prior: Won't this not be compatible with sklearn interfaces? I think there was an idea to pass this with separate `load_model()` and `save_model()` methods proposed by @ggprior

src/tabpfn_client/endpoints/__init__.py
```diff
@@ -0,0 +1,11 @@
+#  Copyright (c) Prior Labs GmbH 2026.
+#  Licensed under the Apache License, Version 2.0
+"""HTTP client for a self-hosted TabPFN inference container.
+
+from tabpfn_client.endpoints import TabPFNClassifier, TabPFNRegressor
```
ggprior: `.endpoints`, not sure -- maybe just `.selfhosted` or `.hosted`?

src/tabpfn_client/hosted/estimator.py
```diff
@@ -0,0 +1,264 @@
...
+"""scikit-learn estimators for a TabPFN inference container reached over HTTP.
+
+Mirrors the `tabpfn_client.TabPFNClassifier` / `TabPFNRegressor` surface;
+each `predict*` call POSTs to the user-supplied `endpoint_url` (the full
+scoring URL, including the `/predict` path). `fit()` does not call the
+endpoint — it just stores `X` / `y` on the estimator. The training data
+is shipped to the endpoint on the next `predict*` call, where the actual
+fit runs.
+
+Auth is optional: when `api_key` is set it is sent as `Bearer <api_key>`,
```
ggprior, line 12: is this accurate? why would we require an api key for self hosted?

src/tabpfn_client/hosted/estimator.py
```diff
@@ -0,0 +1,264 @@
...
+        return np.asarray(result["prediction"])
+
+
+class TabPFNRegressor(_EndpointBase, RegressorMixin):
+    """TabPFN regressor backed by a self-hosted inference endpoint.
+
+    Example:
+        from tabpfn_client.endpoints import TabPFNRegressor
+        reg = TabPFNRegressor(
+            endpoint_url="https://<your-endpoint>/predict",
```
ggprior, line 253: normally we specify the base url here in case we add multiple endpoints later. I'd say in this case, could be even likely?

src/tabpfn_client/endpoints/estimator.py
ggprior: +1 on not messing with sklearn interfaces

ggprior (approved):
I think here we'd have the opportunity to cleanly implement sklearn interfaces, as in server, which makes sense now given kv cache semantics--- fit returns model_id, predict predicts.
Is there a reason to stick with single endpoint here? This was more a workaround for the market places?

otherwise LGTM

simo-prior (approved):
Two points:
1. `model_id` in the `predict()` signature is problematic, it invalidates the reason we (painfully) try to adhere to the sklearn estimator's interface.
2. I'm not sure I buy this should be a separate interface entirely from standard estimators. I think we should rather have a common base class.

Deferring to you on both points, but I think 1) is important.

src/tabpfn_client/hosted/estimator.py
safaricd: There is no base URL in this case, it entirely depends on the customer setup - the best we can do is `http://localhost:8080` which is the default that the container exposes, but having `endpoint_url != None` would most likely lead to user errors, so I'd leave it as required.

src/tabpfn_client/hosted/estimator.py
safaricd: Leaving this as an option so that in the future we can have the Azure and AWS SageMaker marketplace clients inherit from this class - e.g. Foundry requires an API key.

src/tabpfn_client/endpoints/__init__.py
safaricd: Changed - good suggestion.

src/tabpfn_client/endpoints/estimator.py
safaricd: Good point - changed the interface such that it's compatible with other parts of the `sklearn` stack.

src/tabpfn_client/hosted/estimator.py
```diff
@@ -0,0 +1,269 @@
...
+        if self.model_id is not None:
+            body["context"] = {"model_id": self.model_id}
+        else:
+            check_is_fitted(self, ["X_train_", "y_train_"])
+            # y_train on the wire is 2D (n_samples, 1).
+            y_arr = np.asarray(self.y_train_)
+            if y_arr.ndim == 1:
+                y_arr = y_arr.reshape(-1, 1)
+            body["X_train"] = _to_jsonable(self.X_train_)
+            body["y_train"] = y_arr.tolist()
```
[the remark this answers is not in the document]
safaricd: Good catch, resolved ✅

PR #347, opened 2026-07-23
Enable fit_with_cache for predictions on API with K/V

safaricd:
### Change Description

- Enable `fit_with_cache` predictions - previously setting the `fit_mode` was blocked by the API.
- Naming conventions - refactor to be closer with `scikit-learn` standards.

### DIFF
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -53,10 +53,8 @@ class ClassifierPredictParams(BaseModel):
 
 
 class FitMode(str, Enum):
-    LOW_MEMORY = "low_memory"
     FIT_PREPROCESSORS = "fit_preprocessors"
     FIT_WITH_CACHE = "fit_with_cache"
-    BATCHED = "batched"
 
 
 class ClassifierTabPFNConfig(BaseModel):
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -53,11 +53,10 @@ class ClassifierPredictParams(BaseModel):
 
 
 class FitMode(str, Enum):
-    # Only the two values that gapi wires end-to-end. The server-side enum
-    # also has internal `low_memory` / `batched` modes, but those are not
-    # accepted by the public API — see `FitRequest._validate_fit_mode`.
+    LOW_MEMORY = "low_memory"
     FIT_PREPROCESSORS = "fit_preprocessors"
     FIT_WITH_CACHE = "fit_with_cache"
+    BATCHED = "batched"
 
 
 class ClassifierTabPFNConfig(BaseModel):
@@ -126,10 +125,8 @@ class ModelLimit(BaseModel):
     max_classes: int
     max_cols: int
     test_set_max_rows_w_full_regression_output: int
+    predict_row_pairs_budget: int
     test_set_max_cells: int
-    # Optional until all deployed servers send it; estimator.py falls back to
-    # its local default when absent.
-    predict_row_pairs_budget: int | None = None
 
 
 class ModelVersion(str, Enum):
@@ -204,7 +201,6 @@ class DuplicateTestSetErrorResponse(BaseModel):
     message: str
     error_code: str = "DUPLICATE_TEST_SET_UPLOAD"
     trace_id: UUID | None = None
-    detail: str | None = None
     test_set_upload_id: UUID
 
 
@@ -212,7 +208,6 @@ class DuplicateTrainSetErrorResponse(BaseModel):
     message: str
     error_code: str = "DUPLICATE_TRAIN_SET_UPLOAD"
     trace_id: UUID | None = None
-    detail: str | None = None
     train_set_upload_id: UUID
 
 
@@ -268,7 +263,6 @@ class NotFoundErrorResponse(BaseModel):
     message: str
     error_code: str = "NOT_FOUND"
     trace_id: UUID | None = None
-    detail: str | None = None
 
 
 class PredictRequest(BaseModel):
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -53,10 +53,11 @@ class ClassifierPredictParams(BaseModel):
 
 
 class FitMode(str, Enum):
-    LOW_MEMORY = "low_memory"
+    # Only the two values that gapi wires end-to-end. The server-side enum
+    # also has internal `low_memory` / `batched` modes, but those are not
+    # accepted by the public API — see `FitRequest._validate_fit_mode`.
     FIT_PREPROCESSORS = "fit_preprocessors"
     FIT_WITH_CACHE = "fit_with_cache"
-    BATCHED = "batched"
 
 
 class ClassifierTabPFNConfig(BaseModel):
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -125,7 +125,9 @@ class ModelLimit(BaseModel):
     max_classes: int
     max_cols: int
     test_set_max_rows_w_full_regression_output: int
-    predict_row_pairs_budget: int
+    # Optional: older servers don't send this; validate_test_set falls back to
+    # FALLBACK_PREDICT_ROW_PAIRS_BUDGET when absent.
+    predict_row_pairs_budget: int | None = None
     test_set_max_cells: int
 
 
--- a/src/tabpfn_client/api_models.py
+++ b/src/tabpfn_client/api_models.py
@@ -52,6 +52,13 @@ class ClassifierPredictParams(BaseModel):
     ) = None
 
 
+class FitMode(str, Enum):
+    LOW_MEMORY = "low_memory"
+    FIT_PREPROCESSORS = "fit_preprocessors"
+    FIT_WITH_CACHE = "fit_with_cache"
+    BATCHED = "batched"
+
+
 class ClassifierTabPFNConfig(BaseModel):
     n_estimators: int | None = None
     categorical_features_indices: list[int] | None = None
@@ -65,6 +72,7 @@ class ClassifierTabPFNConfig(BaseModel):
         Annotated[Literal["autocast", "auto"] | str, Field(union_mode="left_to_right")] | None
     ) = None
     ignore_pretraining_limits: bool | None = None
+    fit_mode: Annotated[FitMode | UnknownEnum, Field(union_mode="left_to_right")] | None = None
     model_path: str | None = None
     balance_probabilities: bool | None = None
 
@@ -161,6 +169,7 @@ class RegressorTabPFNConfig(BaseModel):
         Annotated[Literal["autocast", "auto"] | str, Field(union_mode="left_to_right")] | None
     ) = None
     ignore_pretraining_limits: bool | None = None
+    fit_mode: Annotated[FitMode | UnknownEnum, Field(union_mode="left_to_right")] | None = None
     model_path: str | None = None
 
 
@@ -226,7 +235,7 @@ class FitRequest(BaseModel):
     )
     async_mode: bool | None = Field(
         default=None,
-        description="Submit the fit and return immediately with status=pending; poll GET /tabpfn/fit/{fitted_train_set_id} for the outcome. Defaults to the blocking behaviour.",
+        description="Submit the fit and return immediately with status=pending; poll GET /fit/{fitted_train_set_id} for the outcome. Defaults to the blocking behaviour.",
     )
 
 
@@ -239,6 +248,7 @@ class FitStatusResponse(BaseModel):
     fitted_train_set_id: UUID
     status: Annotated[FitStatus | UnknownEnum, Field(union_mode="left_to_right")]
     error: str | None = None
+    error_code: str | None = None
 
 
 class GetModelLimitsResponse(BaseModel):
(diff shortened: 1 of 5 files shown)


src/tabpfn_client/api_models.py
```diff
@@ -52,6 +52,13 @@ class ClassifierPredictParams(BaseModel):
     ) = None
 
 
+class FitMode(str, Enum):
+    LOW_MEMORY = "low_memory"
```
safaricd: Even though the API model defines all support fit mode types, it rejects anything but `fit_preprocessors` and `fit_with_cache`

simo-prior: What about removing `LOW_MEMORY` and `BATCHED` anyway? At the moment we don't support them, I feel like it should be a future concern, otherwise this could be confusing for the users.

src/tabpfn_client/estimator.py
```diff
@@ -290,21 +419,29 @@ def __init__(
         self.inference_precision = inference_precision
         self.random_state = random_state
         self.inference_config = inference_config
+        self.fit_mode: FitMode | None = fit_mode
```
simo-prior: Nit: probably not needed to specify the type as it will infer it anyway.

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+def _read_model_handle(handle: dict[str, Any] | str | Path) -> dict[str, Any]:
+    """Accept either an in-memory handle dict or a path to a JSON file
+    produced by `save_model()`."""
+    if isinstance(handle, dict):
+        return handle
+    return json.loads(Path(handle).read_text())
+
+
+def _build_model_handle(
+    estimator: Any,
```
simo-prior: We could have a type for the estimator here.

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+    task: Literal["classification", "regression"],
+    path: str | Path | None,
+) -> dict[str, Any] | Path:
+    """Shared implementation of `save_model()` for both estimators."""
+    check_is_fitted(estimator)
+    params = estimator.get_params(deep=False)
```
simo-prior, line 122: Here you want to use: `estimator._get_tabpfn_config()`, type safe and you get exactly the params that end up in tabpfn (no client_options for instance). Make it public if you prefer.

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+    fitted = getattr(estimator, "fitted_train_set_id_", None)
+    if fitted is not None:
+        return fitted
+    try:
+        return _cast_fitted_id(estimator.model_id)
+    except (ValueError, TypeError):
+        return None
+
+
+def _read_model_handle(handle: dict[str, Any] | str | Path) -> dict[str, Any]:
```
simo-prior: Nit: too much indirection here, I would avoid the extra function.

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+
+def _load_model_handle_for(
+    cls: type,
+    handle: dict[str, Any] | str | Path,
+    task: Literal["classification", "regression"],
+) -> dict[str, Any]:
+    """Read + validate a `save_model()` handle for the target estimator class."""
+    data = _read_model_handle(handle)
+    version = data.get("format_version")
+    if version != _MODEL_HANDLE_VERSION:
```
simo-prior: I would avoid the concept of `_MODEL_HANDLE_VERSION` altogether. The handle can be malformed for a multitude of reasons and in fact this could pass even when `params` is missing. I'd make a pydantic class for the model handle where params is a **TabpfnConfig, and if we fail to deserialize it we give a "Malformed error...possibly no longer support format" + we explain why the deserialization failed.

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+        return handle
+    out = Path(path)
+    out.write_text(json.dumps(handle, indent=2))
+    return out
+
+
+def _load_model_handle_for(
+    cls: type,
+    handle: dict[str, Any] | str | Path,
+    task: Literal["classification", "regression"],
```
simo-prior: PredictionTask from api_models for type safety?

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+    # `client_options` is runtime-only (timeouts, auth/trace headers) and not
+    # JSON-serializable; drop it so the handle stays portable.
+    params.pop("client_options", None)
+    # Persist the *effective* id: a real `fit()` supersedes the constructor's
+    # `model_id`. Coerce to str for JSON.
+    params["model_id"] = str(_resolve_train_set_id(estimator))
+    handle: dict[str, Any] = {
+        "format_version": _MODEL_HANDLE_VERSION,
+        "task": task,
+        "params": params,
```
simo-prior: Nit: since params here is really the `tabpfn_config` obtained with `_get_tabpfn_config()` I'd called it `tabpfn_config`.

src/tabpfn_client/api_models.py
safaricd: Removed ✅

src/tabpfn_client/estimator.py
```diff
@@ -68,6 +71,112 @@
...
+    a resolver instead of syncing state at both mutation sites means `fit()`
+    never has to mutate `self.model_id`, so `get_params()` / `clone()` stay
+    sklearn-correct.
+
+    Returns None if `model_id` is set but not a valid UUID — sklearn's
+    convention is that `__init__` does no validation, so bad ids must not
+    crash `__sklearn_is_fitted__`. The estimator degrades to "not fitted"
+    and `predict` surfaces the standard `NotFittedError`.
+    """
+    fitted = getattr(estimator, "fitted_train_set_id_", None)
```
simo-prior: I don't think we still need `fitted_train_set_id_`, why not always use `model_id`?

src/tabpfn_client/estimator.py
```diff
@@ -277,8 +388,26 @@ def __init__(
...
+            server default, which is "fit_preprocessors".
+            "fit_preprocessors" fits only the preprocessing state, so every
+            predict re-runs the forward pass from the uploaded train set.
+            "fit_with_cache" additionally builds and persists a server-side KV
+            cache keyed by the resulting fitted-train-set id; later predicts
+            against that id (see `model_id` / `save_model` / `load_model`) are
+            served from the cache instead of re-fitting.
         client_options : ClientOptions, default=None
             Client specific options (e.g. timeout, headers).
+        model_id : str, UUID or None, default=None
```
simo-prior: Safe to set the `model_id` directly? Shouldn't it be private and save/load are basically setter and getter?

src/tabpfn_client/estimator.py
simo-prior, line 462: I wonder whether we should change trace_id on `predict_count` > 0?

simo-prior (commented):
@safaricd left comments, let me know.

src/tabpfn_client/estimator.py
safaricd: Done ✅

src/tabpfn_client/estimator.py
safaricd: Done ✅

src/tabpfn_client/estimator.py
safaricd: The reason we use `get_params` is because `_get_tabpfn_config` returns wide-config fields only - it drops `thinking_*`, `model_id` etc., and `save_model` needs the full reproducibility set.

src/tabpfn_client/estimator.py
safaricd: Done ✅

src/tabpfn_client/estimator.py
safaricd: Done ✅

src/tabpfn_client/estimator.py
safaricd: Done ✅

src/tabpfn_client/estimator.py
safaricd: I'd keep the `fitted_train_set_id_` because of sklearn-compatibility. Specifically, if `fit()` writes to `self.model_id`, then `clone(fitted_estimator)` carries the fitted ID into the clone, which violates sklearn's clone contract - which should strip learned state.

src/tabpfn_client/estimator.py
safaricd: I'd keep it private because `sklearn` requires init params to be public attributes with the same name as the constructor args. Making it `_model_id`  would break `get_params()` and `clone()`

src/tabpfn_client/estimator.py
simo-prior: @safaricd but cloning after `load_model()` would equally carry learned state to the new estimator. I think what we want is just one variable but rather a `model_id_` to differentiate from hyperparams. This would solve the cloning problem and also other unexpected behaviors like setting `estimator.model_id` silently not working if we ever called `fit()` on the model.

simo-prior (dismissed):

src/tabpfn_client/estimator.py
```diff
@@ -75,6 +80,110 @@
 
 _VALID_THINKING_EFFORT_LEVELS = frozenset({"medium", "high"})
 
+# `FitMode` exposes only the two values gapi wires end-to-end:
```
ggprior: internal info which is not needed here?

src/tabpfn_client/api_models.py
```diff
@@ -52,6 +52,13 @@ class ClassifierPredictParams(BaseModel):
     ) = None
 
 
+class FitMode(str, Enum):
```
ggprior, line 55: why do we declare enums like low_memory? clients cannot set that

src/tabpfn_client/estimator.py
```diff
@@ -305,13 +430,15 @@ def __init__(
...
         self._last_trace_id = None
-        self._last_fitted_train_set_id = None
         self._last_train_X = None
         self._last_train_y = None
         self._last_meta = {}
         self._last_train_set_description = None
         self._fit_count = 0
 
+    def __sklearn_is_fitted__(self) -> bool:
+        return getattr(self, "model_id_", None) is not None
```
ggprior, line 435: is this actually true (even if you don't fit with cache, is this parameter set correctly)

src/tabpfn_client/estimator.py
```diff
@@ -851,6 +1066,39 @@ def _validate_targets(self, y) -> None:
         if sum(pd.isnull(y_)) > 0:
             raise ValueError("Input y contains NaN.")
 
+    @overload
+    def save_model(self, path: None = None) -> dict[str, Any]: ...
+    @overload
+    def save_model(self, path: str | Path) -> Path: ...
```
ggprior, line 1067: pretty cool! but isn't it conflicting with the server-side caching/dedup logic? Shouldn't we remove that first and then add save/load semantics afterwards?

src/tabpfn_client/estimator.py
```diff
@@ -305,13 +430,15 @@ def __init__(
...
 
         self._last_trace_id = None
-        self._last_fitted_train_set_id = None
         self._last_train_X = None
         self._last_train_y = None
         self._last_meta = {}
         self._last_train_set_description = None
         self._fit_count = 0
 
+    def __sklearn_is_fitted__(self) -> bool:
```
ggprior, line 434: Why do we need the is fitted function here, it would be perfectly valid with underscore attribute convention as well
https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html#sklearn.utils.validation.check_is_fitted

src/tabpfn_client/estimator.py
ggprior: I feel this is overloading this PR to be honest. I.e. save/load semantics are a separate concern from adding KV cache support, and should be tracked separately.
I think this needs a lot more work in terms of testing before shipping something here

ggprior (changes requested):
LGTM mostly, 
ignore the stuff about save/load model -- if it's already there, fine
I think we should address leakage of server side internals
Also the fit_mode which looks auto-generated
Then this is GTG

src/tabpfn_client/estimator.py
simo-prior: addressed

src/tabpfn_client/estimator.py
simo-prior: It's set on every fit, see: https://github.com/PriorLabs/tabpfn-client/pull/347/changes/BASE..9b85e7285c23d57535274f6567a9f42508e996d0#diff-33bad89c7747dd680e3135ec5d350edb8949a3e1452e672861f1dbb6a5ed9856R491

src/tabpfn_client/estimator.py
simo-prior: On the classifier `classes_` is set before the remote fit call succeeds, so `model_id_` might never get populated. I don't think the regressor needs it today, but it doesn't hurt to keep it on both for consistency.

ggprior:
Superseded by #351

safaricd:
Closing this one because `fit_with_cache` was implemented here - https://github.com/PriorLabs/tabpfn-client/pull/351

PR #348, opened 2026-07-24
Adaptive per-call test set row limit at predict time

klemens-floege:
## What

The API now accepts much larger test sets, and the static row/cell limits the client enforces already come from `get_model_limits`. What was missing client-side is the adaptive part: prediction compute scales with `n_train_rows * n_test_rows`, so the effective per-call test row limit shrinks as the fitted training set grows.

- Add `PREDICT_ROW_PAIRS_BUDGET` (250k × 1M): at predict time the test row limit is `min(test_set_max_rows, budget // n_train_rows)` — e.g. 1M training rows → 250k test rows per call, ≤ 250k training rows → the static 1M cap binds.
- `validate_test_set` takes an optional `train_rows`; both `TabPFNClassifier` and `TabPFNRegressor` pass the fitted training set size, so oversized test sets fail fast client-side with a clear "split across multiple calls" message instead of erroring after upload.

| train rows | test row limit per call |
|---|---|
| 1M | 250k |
| 500k | 500k |
| ≤ 250k | 1M (static cap) |

## Tests

Unit tests cover the static cap, the adaptive limit at the 1M-train anchor, the adaptive limit never exceeding the static cap, and the offline no-op when limits are unavailable.

### DIFF
--- a/src/tabpfn_client/estimator.py
+++ b/src/tabpfn_client/estimator.py
@@ -66,6 +66,11 @@ _AUTO_MODEL_PATH_ALIASES: frozenset[str | None] = frozenset({None, "auto", "defa
 
 THINKING_TIMEOUT_MAX_S = 40 * 60
 
+# Prediction compute scales with n_train_rows * n_test_rows, so the API caps
+# their product. The effective per-call test row limit therefore shrinks as
+# the fitted training set grows (e.g. 1M training rows -> 250k test rows).
+PREDICT_ROW_PAIRS_BUDGET = 250_000 * 1_000_000
+
 _VALID_THINKING_EFFORT_LEVELS = frozenset({"medium", "high"})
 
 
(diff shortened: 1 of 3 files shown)


src/tabpfn_client/estimator.py
```diff
@@ -66,6 +66,11 @@
 
 THINKING_TIMEOUT_MAX_S = 40 * 60
 
+# Prediction compute scales with n_train_rows * n_test_rows, so the API caps
+# their product. The effective per-call test row limit therefore shrinks as
+# the fitted training set grows (e.g. 1M training rows -> 250k test rows).
+PREDICT_ROW_PAIRS_BUDGET = 250_000 * 1_000_000
```
simo-prior: @klemens-floege I was thinking we probably want the server to send this down as well right? For instance the 1M limit is likely to change. To do that it would need to be added here: https://github.com/PriorLabs/tabpfn-server/pull/1037/changes#diff-590aed662160a700fba727922db79442aef09c6821e3e29db8481e4af882bbfcR55 and `api_models.py` should be updated accordingly.

klemens-floege: please see: https://github.com/PriorLabs/tabpfn-server/pull/1076/

simo-prior (approved):

PR #351, opened 2026-07-30
Support async fits via submit job

simo-prior:
[ENG-810](https://linear.app/priorlabs/issue/ENG-810/support-async-fits-via-submit-job)

**Description**

- [x] Use new endpoints: `api_settings` (from model_limits), `submit_fit_job`.
- [x] Add ApiMode options (sync, async and auto)
- [x] Implement server guidance on polling via `retry_in_secs`
- [x] Remove NeedsRefitting (possible after v0-removal)
- [x] Better user-friendly warnings
- [x] Refactor `_validate_response()`


src/tabpfn_client/client.py
```diff
@@ -409,77 +402,73 @@ def fit(
...
                 headers=client_options.headers,
             )
+        else:
+            fit_resp = cls._fit(
+                req=FitRequest(
+                    train_set_upload_id=prepare_resp.train_set_upload_id,
+                    task_config=task_config,
+                    tabpfn_systems=tabpfn_systems_,
+                    thinking_config=thinking_config,
+                ),
```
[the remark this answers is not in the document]
simo-prior: Wrong finding, None is accepted.

src/tabpfn_client/estimator.py
```diff
@@ -320,23 +351,26 @@ def fit(
...
-            self.thinking_metric,
-            self.model_path,
-        )
         X_clean = _clean_text_features(X)
-        self._validate_targets_and_classes(y)
+        classes = self._validate_targets_and_classes(y)  # XXX
+
+        self.thinking_mode = _resolve_thinking_mode(
+            self.thinking_mode, self.thinking_effort
+        )
```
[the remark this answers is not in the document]
simo-prior: Addressed

src/tabpfn_client/client.py
```diff
@@ -409,113 +401,172 @@ def fit(
...
+            res,
+            "get_fit_status",
+            success_model=GetFitStatusResponse,
+        )
+
+    @classmethod
+    # Job submission is not idempotent (each accepted request enqueues a new
+    # fit job), so only connection-setup failures — where the request never
+    # reached the server — are safe to retry. Anything after send (read
+    # timeouts, 5xx) may mean the job was already enqueued, and a blind retry
```
brendan-priorlabs: This is not true depending on where the 5xx happens. I'm not sure this is particularly robust. Probably the client generating a random request ID and allowing the server to de-dupe with that could help here. Not sure what the best course of action would be without a server change

src/tabpfn_client/client.py
```diff
@@ -809,156 +849,103 @@ def _warn_if_deprecated(response: httpx.Response) -> None:
...
-            load = response.json()
-        except json.JSONDecodeError as e:
-            logging.info(f"Failed to parse JSON from response in {method_name}: {e}")
-
-        # Streamed-error envelope: long-running endpoints (e.g. /tabpfn/fit with
-        # thinking mode) return a chunked 200 response, and on mid-stream failure
-        # the server emits `{"_streamed_error": True, "message": "..."}` as the
-        # body. Without this short-circuit, the success-schema validation below
-        # turns that into a misleading pydantic `Field required` error and the
-        # real message is buried inside `input_value`.
```
brendan-priorlabs, line 840: Remind me why this is safe now?

src/tabpfn_client/estimator.py
```diff
@@ -284,6 +294,20 @@ def __init__(
                 "roc_auc_ovr_micro", "roc_auc_ovr_weighted".
 
             Aliases "acc", "nll", "pac_score" are also accepted.
+        fit_mode: {"fit_preprocessors", "fit_with_cache"} or None, default=None
```
brendan-priorlabs: Nit: order as in constructor

src/tabpfn_client/estimator.py
```diff
@@ -320,51 +350,53 @@ def fit(
...
+        tabpfn_systems = _build_tabpfn_systems(self.paper_version, thinking_mode)
+        thinking_config = _build_thinking_config(
+            enabled=thinking_mode,
+            effort=self.thinking_effort,
+            timeout_secs=self.thinking_timeout_s,
+            metric=self.thinking_metric,
+        )
 
         if Config.use_server:
+            # NOTE(@trace_id)
```
brendan-priorlabs, line 371: Was this supposed to be a todo of some kind?

src/tabpfn_client/estimator.py
```diff
@@ -852,59 +873,6 @@ def _validate_targets(self, y) -> None:
             raise ValueError("Input y contains NaN.")
 
 
-def _is_thinking_supported_model_path(model_path: str | None) -> bool:
```
brendan-priorlabs, line 855: Why is all this removed? Seems relevant. Maybe I missed it moving elsewhere

src/tabpfn_client/client.py
```diff
@@ -409,113 +401,172 @@ def fit(
...
-                )
-
-            time.sleep(min(get_opts().TABPFN_CLIENT_POLL_INTERVAL, remaining))
+        single_poll_timeout = 15.0
+        async_settings = cls._resolve_async_settings()
+        total_timeout = async_settings.poll_timeout_secs
+        deadline = time.monotonic() + total_timeout
+        retry_in_secs = 5.0  # fallback until the server sends retry_in_secs
 
+        while True:
```
brendan-priorlabs, line 482: Here's one from Claude. Supposedly we're remapping certain hard failures into retryable ones and then catching them. If true, this deserves a fix.

>   3. _wait_for_fit turns hard failures into a 2-hour hang with the real error
>   discarded
> 
>   client.py:469-503
> 
>   The except branch passes and loops until poll_timeout_secs (default 7200s),
>   then raises TimeoutError: ... did not reach a terminal state within 7200
>   seconds — with the actual cause thrown away. Since _get_fit_status now also
>   maps 429 and 500 to RetryableServerError (client.py:525), these all become a
>   silent 2-hour poll: server down (NetworkError), a deterministic 500, or a
>   quota/rate-limit 429 during polling.

brendan-priorlabs (commented):
Thanks for this change, @simo-prior. One general thing I think is worth considering for this sort of PR is scope. I think the critical changes are a subset of what's here. Granted, we have to balance cleaning up as we go with accumulating tech debt, but the larger scope increases the probability of regressions and makes reviewing more time consuming as well (or less effective with a lighter review). Something to keep in mind for the next one. Thanks again for tackling this!

src/tabpfn_client/client.py
simo-prior: Agreed, the comment needs some rephrasing. The point is we need to be more careful retrying fits now that we don't de-dup them. For now I think the logic is correct, we just fail and let the user retry. Re-phreasing the comment.

src/tabpfn_client/client.py
simo-prior: It's not, in fact it's just moved here: https://github.com/PriorLabs/tabpfn-client/pull/351/changes#diff-a9da9e7843c9a0e2c9b17fd7589b98aacce7f55ef9c6fb5b306a08c7eca3cd4eR930

The idea being reducing `_read_json_body` responsability to just parsing the body.

src/tabpfn_client/estimator.py
simo-prior: addressed

src/tabpfn_client/estimator.py
simo-prior: No just my way of searching for tags in the code 😅

src/tabpfn_client/estimator.py
simo-prior: Yes this is fully moved to the server.

src/tabpfn_client/client.py
simo-prior: I think `429` is safe to retry, especially since we are already backing off. I missed 500 indeed, removing it 👍 

Also making sure error doesn't get swalloed by TimeoutError.

brendan-priorlabs: 429 is how we signal quota exhausted, right? I don't think that's safe at all. Needs to be surfaced to the user so they can request more quota

brendan-priorlabs (dismissed):
LGTM. Left one final followup comment that I think is very important to the user experience, so please be sure that that is definitely fixed before merging. Thanks a bunch, @simo-prior!

brendan-priorlabs (approved):

PR #352, opened 2026-08-10
Token-only auth by default, with opt-in interactive_login()

ggprior:
The default path no longer contains "magic" to prompt for tokens. Instead we explicitly require a user to pass a token.

Error message clearly describes what to do---

Screenshot 2026-08-20 at 15 30 18

Interactive Login is behind `tabpfn_client.interactive_login() only` --

Screenshot 2026-08-20 at 15 32 05


ggprior:
## Follow-up: interactive signup flow

Feedback on the signup steps (role list, redundant steps, email opt-in wording) is tracked separately in **[OPC-26 — Rework interactive signup flow in tabpfn-client](https://linear.app/priorlabs/issue/OPC-26/rework-interactive-signup-flow-in-tabpfn-client)**.

It is deliberately out of scope for this PR, which is about *where* authentication happens, not what the signup form asks. OPC-26 carries three open questions that need answers before anyone implements — notably whether a CLI-native signup should exist at all now that registration happens in the browser.

src/tabpfn_client/interactive_auth.py
```diff
@@ -0,0 +1,388 @@
...
+    sys.stdout.write(prompt)
+    sys.stdout.flush()
+
+    waited = 0.0
+    while not auth_event.is_set():
+        if waited >= timeout:
+            sys.stdout.write("\nTimed out waiting for login.\n")
+            sys.stdout.flush()
+            return None
+        ready, _, _ = select.select([sys.stdin], [], [], 0.5)
```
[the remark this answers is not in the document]
safaricd: @ggprior should we address this Windows incompatibility issue?

src/tabpfn_client/constants.py
```diff
@@ -6,7 +6,9 @@
 CACHE_DIR = Path(__file__).parent.resolve() / ".tabpfn"
 
 URL_TABPFN_CLIENT_GITHUB_ISSUES = "https://github.com/priorlabs/tabpfn-client/issues"
+URL_PRIOR_LABS_API_KEYS = "https://ux.priorlabs.ai/account/api-keys"
```
safaricd, line 9: Can we have this configurable via env vars with https://ux.priorlabs.ai/account/api-keys as the default? The reason this would be convenient to have is using the client on the staging environment.

src/tabpfn_client/prompt_agent.py
```diff
@@ -5,24 +5,41 @@
...
     status,
+    success,
+    warn,
     print_logo,
 )
 
+# The registration link the server expects; kept from the original flow.
+VALIDATION_LINK = "tabpfn-2023"
+
+ROLES = [
```
safaricd, line 31: Out of scope for this PR, but would be nice to have these retrieved from the API instead of statically declared in-code to avoid future drift.

src/tabpfn_client/service_wrapper.py
```diff
@@ -87,32 +111,43 @@ def set_token_by_registration(
...
-        if ServiceClient.get_access_token() is None:
-            if get_opts().TABPFN_TOKEN:
-                access_token = get_opts().TABPFN_TOKEN
-            else:
-                if not cls.CACHED_TOKEN_FILE.exists():
-                    return False, None
-                access_token = cls.CACHED_TOKEN_FILE.read_text()
-        else:
-            access_token = ServiceClient.get_access_token()
+        access_token = cls.resolve_token()
```
safaricd: Here's one edge case I got locally:
1. Clean install `tabpfn_client`
2. Run `from tabpfn_client import interactive_login` and `interactive_login()`
3. Run a classifier or regressor, success ✅ 
4. Set `os.environ["TABPFN_TOKEN"] = "bogus"` to an invalid token
5. Run `TabPFNClassifier.fit` -> `init` -> `try_reuse_existing_token`
6. `_reset_token()` calls `CACHED_TOKEN_FILE.unlink(missing_ok=True)` and deletes the good cached token. 
7. Unset the env var, retried - cache is gone, must run `interactive_login()` again.

Not critical, but a path worth considering.

src/tabpfn_client/interactive_auth.py
```diff
@@ -0,0 +1,388 @@
...
+                    stays available throughout.
+    :returns: The access token.
+    :raises InteractiveLoginError: If no terminal is available, the flow was
+                                   aborted, or the resulting token was rejected.
+    """
+    if not force_relogin:
+        from tabpfn_client.service_wrapper import UserAuthenticationClient
+
+        existing = UserAuthenticationClient.resolve_token()
+        if existing is not None and ServiceClient.is_auth_token_outdated(existing):
```
safaricd: `ServiceClient.is_auth_token_outdated` runs a HTTP GEt with no `try/except` around the `httpx` call. If our API/server is unreachable from the local network, we display the `httpx.ConnectError` traceback instead of falling back to the auth flow.

src/tabpfn_client/interactive_auth.py
```diff
@@ -0,0 +1,388 @@
...
+                f"{body}</body></html>"
+            )
+            self.wfile.write(html.encode())
+            if received_token[0] is not None:
+                auth_event.set()
+
+        def log_message(self, format: str, *args: object) -> None:
+            pass  # silence request logs
+
+    httpd = socketserver.TCPServer(("", 0), _CallbackHandler)
```
safaricd: This automatically binds to `0.0.0.0` which may not be secure, consider switching to `127.0.0.1`

src/tabpfn_client/interactive_auth.py
```diff
@@ -0,0 +1,388 @@
...
+
+    callback_url = f"http://localhost:{port}"
+    login_url = f"{gui_url}/login?callback={callback_url}"
+
+    server_thread = threading.Thread(
+        target=_serve_until_event, args=(httpd, auth_event), daemon=True
+    )
+    server_thread.start()
+
+    webbrowser.open(login_url)
```
safaricd: Shouldn't we check here for the return value of `webbrowser.open`? In case opening a browser failed, the user has to wait for 10 minutes before acknowledging the error.

src/tabpfn_client/constants.py
ggprior: can do but note this only controls the log message, client staging controls are untouched here

safaricd: In that case, let's leave it as is - thanks for the heads up

src/tabpfn_client/interactive_auth.py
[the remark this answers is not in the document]
ggprior: fixed

src/tabpfn_client/interactive_auth.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+            return existing
+
+    if not _stdin_is_interactive():
+        raise InteractiveLoginError(
+            "interactive_login() needs an interactive terminal.\n"
+            f"Generate a token at {URL_PRIOR_LABS_API_KEYS} and either\n"
+            "  - set it as the TABPFN_TOKEN environment variable, or\n"
+            "  - pass it to tabpfn_client.set_access_token('<your-token>')."
+        )
```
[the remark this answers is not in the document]
ggprior: fixed

src/tabpfn_client/service_wrapper.py
ggprior: fixed now

src/tabpfn_client/interactive_auth.py
ggprior: wrapped

src/tabpfn_client/interactive_auth.py
ggprior: checking now, but note that the paste prompt is done concurrently with webbrowser.open (like say login to Claude) so that concern shouldn't hold -- still fixed and adding explicit check

safaricd (approved):
LGTM, thanks for applying the changes

PR #355, opened 2026-08-17
Add Azure AI Foundry TabPFN client with thinking-mode support

safaricd:
Introduces the Azure AI Foundry TabPFN client (`TabPFNClassifier` / `TabPFNRegressor`) with thinking-mode support and hosted-client-compatible `predict` output shapes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "tabpfn-client"
-version = "0.4.1"
+version = "0.4.2"
 requires-python = ">=3.10"
 
 description = "API access for TabPFN: Foundation model for tabular data"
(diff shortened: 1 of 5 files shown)


ggprior (dismissed):
LGTM, but do you have an exmaple of the inconsistency that was reported, i.e. a diff in return type before & after this change? Hard to tell otherwise

safaricd:
@ggprior here's what was reported by Bosch:
> The issue was with .predict of tabpfn-client, specific to foundry. In standard tabpfn result = model.predict(Xtest, output_type="full") returns a result variable which is a dictionary with keys like "mean", "median", etc.
> In tabpfn-client, they do something weird, if I call result = model.predict(Xtest, output_type="full"), the result is a 0-dim numpy ndarray, with a dictionary as the only element. Internally they are doing return np.asarray(result["prediction"]) inside the predict function!
> I then have to do result = result.item() to recover back the standard dictionary form and then continue as usual.
> Also in the tabpfn-client, the criterion variable is missing. It is used, for the bar distribution plot, to compute the variances, like you had showed me. See below the output dictionary for tabpfn3-client and local tabpfn2.2

ggprior (approved):

PR #361, opened 2026-08-24
ci: add release pipeline and towncrier changelog

ggprior:
AI;DR: adding release pipelines following our other repositories including TabPFN, tabpfn-time-series, etc.

<details>
Adds the release pipeline this repo has been missing, matching the shape used in the other
public repos, plus towncrier changelogs so every release gets real notes.

## Flow

| Workflow | Trigger | Does |
|---|---|---|
| `release-create-pr.yml` | manual dispatch (`version`, `base_ref`) | bumps `[project].version`, runs `towncrier build`, opens a `Release vX.Y.Z` PR |
| `release-tag-on-merge.yml` | merge of a `release/v*` branch | reads the version back out of `pyproject.toml` and pushes tag `vX.Y.Z` |
| `release-publish.yml` | `push` of a `v*` tag | verifies tag == pyproject, builds, TestPyPI → install-verify → PyPI → GitHub release |
| `check-changelog.yml` | every PR | requires a `changelog/<PR>.<category>.md` fragment or the `no changelog needed` label |

Publishing to PyPI sits behind the `pypi` environment, so it waits for a review before
anything reaches the real index. TestPyPI runs unguarded ahead of it, and the release is
installed back out of TestPyPI and checked before the gate is reached.

## Differences from the sibling repos

These are deliberate rather than drift:

- **Every action pinned by commit SHA**, including `pypa/gh-action-pypi-publish`. Two of the
  sibling repos carry a comment saying that action cannot be SHA-pinned because it resolves a
  Docker image per ref. That is no longer true — it has been pinned at `dc37677b` elsewhere
  since 2026-07-30 and has published several releases through it. Every SHA here was checked
  against its tag via the API rather than copied on trust. Dependabot already tracks
  `github-actions` weekly, so the pins stay current.
- **No `${{ }}` interpolation inside `run:` blocks.** Dispatch inputs and PR head refs reach
  the shell through `env:` and are quoted, and each is validated against a pattern before
  first use. Splicing them directly into a script is a template-injection path.
- **`permissions: {}` at workflow level**, granted per job — the publish job is the only one
  that gets `contents: write`.
- **`--prerelease` on the GitHub release** when the version is a PEP 440 pre-release, so an
  rc does not show up as the latest release.
- **The release PR is opened with `RELEASE_BOT_TOKEN`.** A PR opened with the `GITHUB_TOKEN`
  does not start workflow runs, which would leave the release PR with no CI at all.
- `--verify-tag` on release creation, `persist-credentials: false` on checkouts that do not
  push, and a `concurrency` group per tag.

## Also here

- `tabpfn_client.__version__`, resolved from installed package metadata. The publish job
  installs from TestPyPI and asserts the version matches what it built; there was nothing to
  assert against before.
- Retroactive changelog fragments for #352 and #359, which merged before this existed. Without
  them the first `towncrier build` would find nothing and the release workflow would stop.
- `CHANGELOG.md` seeded with an `## [Unreleased]` heading, and `changelog/README.md`
  documenting how to add a fragment.

## Setup this needs before a release can run

1. Trusted publishers on PyPI and TestPyPI for `tabpfn-client` — repo `PriorLabs/tabpfn-client`,
   workflow `release-publish.yml`, environments `pypi` and `testpypi`. No API token is used.
2. Environments `release`, `testpypi` and `pypi`, with a required reviewer on `pypi`.
3. `RELEASE_BOT_TOKEN`, a fine-grained PAT scoped to this repo with `contents: write` and
   `pull-requests: write`. A tag pushed with the `GITHUB_TOKEN` does not start a workflow run,
   so without it the publish workflow never fires.

Verified locally: `actionlint` is clean on all four workflows, `towncrier build` produces the
expected `CHANGELOG.md` from the seeded fragments, the notes-extraction and version regexes
were run against the real files, and the unit suite still passes.
</details>

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -11,10 +11,15 @@ permissions: {}
 jobs:
   check:
     name: Changelog entry
+    # The release-branch exemption applies to this repository's own branches
+    # only. `head_ref` is the head branch's name wherever it lives, so a fork
+    # branch called release/v... would otherwise skip the check -- and a
+    # skipped required check reports as passing.
     if: >-
       github.event_name == 'pull_request' &&
       github.actor != 'dependabot[bot]' &&
-      !startsWith(github.head_ref, 'release/v')
+      !(github.event.pull_request.head.repo.full_name == github.repository &&
+        startsWith(github.head_ref, 'release/v'))
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -1,0 +1,57 @@
+name: Check Changelog
+run-name: "changelog check on ${{ github.head_ref || github.ref_name }}"
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled]
+  merge_group:
+
+permissions: {}
+
+jobs:
+  check:
+    name: Changelog entry
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.actor != 'dependabot[bot]' &&
+      !startsWith(github.head_ref, 'release/v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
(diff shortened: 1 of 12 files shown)


ggprior:

- #361 ◀

simo-prior (approved):

PR #374, opened 2026-09-02
Add `v3.5_default` model alias

simo-prior:
[ENG-975](https://linear.app/priorlabs/issue/ENG-975/update-client-for-tabpfn-35-checkpoint-naming)

### Change Description

Adds the `v3.5_default` alias and stops resolving model names on the client: `model_path` now goes to the server unchanged, and the server owns the name → checkpoint mapping.

- `model_path` is sent as given (`v3.5_default`, `v2.5_real`, `gn2p4bpt`, or a full checkpoint filename). `_model_name_to_path`, the family identifier constants and client-side name validation are removed. `None` sends no `model_path`; `"auto"` / `"default"` go through as given and the server resolves them. The client keeps the alias list only to pick the default-version limits for its pre-flight checks.
- The `<version>_default` entries of `list_available_models()` are derived from `ModelVersion` (newest first) instead of hand-maintained constants; `ModelVersion.V3_5` is added. `create_default_for_version` builds the alias the same way and rejects versions not in the enum. The substring-safe ordering that tabpfn-time-series relies on is kept and tested.
- `model_version_from_path` accepts short names as well as filenames, so the client-side pre-flight limit checks and full-output chunking keep working; bare hash names resolve to v2.

Tests cover the pinned model lists, pass-through of the name on the wire and `create_default_for_version(ModelVersion.V3_5)`; tests for the removed helpers are dropped.

Motivation: the tabpfn-server v3.5 smoke test (`basic_v3_5` preset) uses this alias; the pinned client rejects it today.

**If you used new dependencies: Did you add them to `requirements.txt`?** No new dependencies.

**Who did you ping on Mattermost to review your PR?** —

## Breaking changes

**Does this PR break the API? If so, what is the corresponding server commit?** Wire change: the client sends short model names instead of `tabpfn-<family>-<task>-<name>.ckpt` filenames. gapi's `fill_or_validate_model_path` must accept and resolve them (not yet on tabpfn-server `main`). Until it does, any explicit `model_path`, including `"auto"` / `"default"`, fails at fit with the server's invalid-model-path error; only `None` and full checkpoint filenames keep working.

**Does this PR break the user interface? If so, why?** Two small behaviour changes: `create_default_for_version` no longer sets `n_estimators=8` / `softmax_temperature=0.9`, it only pins `model_path`; and unknown model names are no longer rejected locally but by the server at fit time.

### DIFF
--- a/tests/unit/test_tabpfn_regressor.py
+++ b/tests/unit/test_tabpfn_regressor.py
@@ -25,7 +25,7 @@ from tabpfn_client.client import (
     PredictionResult,
     ServiceClient,
 )
-from tabpfn_client.api_models import RegressorTabPFNConfig
+from tabpfn_client.api_models import ModelVersion, RegressorTabPFNConfig
 
 
 def _api_settings_payload(
@@ -857,6 +857,7 @@ class TestTabPFNModelSelection(unittest.TestCase):
 
     def test_list_available_models_returns_expected_models(self):
         expected_models = [
+            "v3.5_default",
             "v3_default",
             "v2.6_default",
             "v2.5_default",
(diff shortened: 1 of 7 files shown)


tests/unit/test_tabpfn_regressor.py
```diff
@@ -925,10 +926,24 @@ def test_model_name_to_path_returns_expected_path(self):
             expected_specific_path,
         )
 
+        # "v3" is a prefix of "v3.5": each has to land in its own family.
+        self.assertEqual(
+            TabPFNRegressor._model_name_to_path("regression", "v3.5_default"),
+            "tabpfn-v3.5-regressor-v3.5_default.ckpt",
```
brendan-priorlabs: I think I @mentioned you on the thread, but this is not the name that we'll be using, right? Did something change?

brendan-priorlabs (changes requested):

ggprior (dismissed):

brendan-priorlabs (approved):

PR #376, opened 2026-09-03
Persist fitted model IDs across script reruns

simo-prior:


ggprior (approved):
Didn't test this, but this looks right. Very nice addition! 👍

PR #378, opened 2026-09-04
Fix invalid inference config values that were passed down to the model

safaricd:
### Change Description

The `tabpfn_client.hosted` had invalid default inference configs. This PR streamlines them with the API's, in which the API (be it the inference container) decides on the configuration that is used, but the user may still overwrite it.


ggprior (dismissed):
funny how such a simple change can cause such churn 🫠

jmkuebler (approved):
stamping

PR #379, opened 2026-09-04
Forward `model_version` to the endpoint

safaricd:
Select a baked weights family (e.g. "v3") when the endpoint supports it. Omitted from the request body when unset, so older containers that reject unknown fields keep working.


ggprior (dismissed):
what happens if we predict against a dist which doesn't have the model?

safaricd:
> what happens if we predict against a dist which doesn't have the model?

@ggprior this is handled and a `4xx` is returned - already tested on a live container.

simo-prior (approved):

PR #380, opened 2026-09-04
Release v0.5.2

ggprior:
Automated release PR for v0.5.2, cut from 61b2f4a76cdf376727cbe65a3739a5ead8fae316.

Merging this tags `v0.5.2` and starts the publish workflow, which
uploads to TestPyPI, installs the result from there to verify it, and
then publishes to PyPI.

The PyPI step runs in the `pypi` environment. Whether it pauses for a
review depends on that environment carrying a required reviewer, which
is repository configuration rather than anything this workflow can
assert -- an environment with no protection rules publishes
unattended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


safaricd (approved):
LGTM

PR #385, opened 2026-09-11
fix: stop printing from fit/predict and demote routine log records

ggprior:
`fit()` and `predict()` wrote a `\r` spinner to stdout on every call, and routine events were logged at WARNING, which Python prints to stderr even when the application has configured no logging. In scripts and CI logs the spinner frames end up in captured output; in notebooks the spinner and log lines share one output line. A library should leave stdout and logging configuration to the application.

- **Spinner removed.** `fit()` and `predict()` call the server directly, without the worker thread the spinner needed. `TABPFN_CLIENT_CI_MODE`, which only disabled the spinner, is removed; setting it is now a no-op.
- **"Train/Test set already exists, skipping upload."** is logged at DEBUG. It is the normal dedup path, e.g. predicting on the same `X` twice.
- **`init()` no longer prints "Found existing access token, reusing it for authentication."** on every run in a terminal. `try_reuse_existing_token()` already logs the same event at DEBUG.
- **`try_connection()`** no longer logs an ERROR and calls `traceback.print_exc()` before `init()` raises its own connection error. The traceback is logged at DEBUG.
- **`UserDataClient`** logs through its module logger. It called `logging.info()` / `logging.error()`, which run `basicConfig()` when the root logger has no handlers, so a later `basicConfig()` in the application did nothing. The error logs that preceded a re-raise are dropped.
- **Missing `tabpfn` / `torch` for `output_type="full"`** is now a `UserWarning` via `warnings.warn`: the caller can act on it and filter it.
- **Unused `ui` helpers** are removed: `setup_logging()` (a `basicConfig()` helper), `header()`, `info()`, `progress_bar()` and `print_logo_small()`. None had callers.

Unchanged: the interactive login and signup flows still print, since they are opt-in and need a terminal, and server greeting messages still go through `notify()`.

Not in this PR: the import-time `setLevel(WARNING)` on the `httpx` / `httpcore` loggers, and the server deprecation notice being a `DeprecationWarning`, which is hidden by default outside `__main__`.

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018bhEkFCJWSiwyStDihRFMa

### DIFF
--- a/changelog/385.changed.md
+++ b/changelog/385.changed.md
@@ -1,0 +1,1 @@
+`fit()` and `predict()` no longer print a progress spinner to stdout, `init()` no longer announces that it is reusing a cached token, and routine messages such as "Test set already exists, skipping upload." are logged at DEBUG instead of WARNING. `TABPFN_CLIENT_CI_MODE`, which only disabled the spinner, is removed.
--- a/src/tabpfn_client/client.py
+++ b/src/tabpfn_client/client.py
@@ -15,7 +15,6 @@ import numpy as np
 import re
 import struct
 import time
-import traceback
 import warnings
 from pydantic import BaseModel, ValidationError
 from typing import Any, cast, Mapping, NoReturn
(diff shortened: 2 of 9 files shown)


ggprior:

- #385 ◀

brendan-priorlabs (approved):

PR #382, opened 2026-09-07
Add API-backed dataset cost estimates

dianaprior:
### Change Description

Adds `estimate_cost(X_train, X_test, ...)` so callers can check an operation's quota cost before running it. The helper sends only dataset dimensions and settings to `POST /tabpfn/estimate_cost`, returning a typed result with `estimated_cost`, `pricing_version`, and the resolved inputs. It does not upload feature values or consume quota.

Supports prediction, Thinking fit/predict, and cached prediction. Defaults are resolved by the server; cache fallback or different fitted estimator counts can change the final charge. The endpoint requires quota v3 to be enabled.

Server counterpart: https://github.com/PriorLabs/tabpfn-server/pull/1164 (endpoint commit `78435846`).

```python
from tabpfn_client import estimate_cost

quote = estimate_cost(X_train, X_test, model_version="v3")
print(quote.estimated_cost, quote.pricing_version)
```

Validation: all 275 client unit tests passed, including 17 new helper tests. Ruff, type checks, and direct client-to-server HTTP contract checks also passed locally.

**New dependencies:** None.

**Mattermost reviewer ping:** None; this PR is a draft.

## Breaking changes

**Does this PR break the API?** No. The new helper requires the server counterpart above; existing client calls are unchanged.

**Does this PR break the user interface?** No.

PR #386, opened 2026-09-11
Release v0.6.0

ggprior:
Automated release PR for v0.6.0, cut from a85960ac0776a53334d2453f67b8beba41c0a74a.

Merging this tags `v0.6.0` and starts the publish workflow, which
uploads to TestPyPI, installs the result from there to verify it, and
then publishes to PyPI.

The PyPI step runs in the `pypi` environment. Whether it pauses for a
review depends on that environment carrying a required reviewer, which
is repository configuration rather than anything this workflow can
assert -- an environment with no protection rules publishes
unattended.


brendan-priorlabs (approved):

ggprior:
We'll do this Monday to also get #382 .

PR #387, opened 2026-09-14
feat: add group_col, time_col and group_time_col for thinking mode

ggprior:
Adds `group_col`, `time_col` and `group_time_col` to `TabPFNClassifier` and `TabPFNRegressor`. They name columns of `X` that describe how rows relate. During a thinking fit, rows of one group stay on one side of every validation split, `time_col` makes validation use contiguous blocks of time, and `group_time_col` orders the rows within a group.

- The columns are sent as new fields of `ThinkingConfig`. Unset fields are dropped from the request as before, so a fit that does not use them sends the same payload.
- `fit` and `predict` check them before any upload: thinking mode is required, `group_col` and `time_col` cannot be combined, `group_time_col` requires `group_col`, and `X` must be a pandas DataFrame that holds the named columns.
- The constructor docstrings (new since 0.6.0) and the README describe them.

Using the new parameters needs a server version that accepts these fields. An older server rejects the request.

https://claude.ai/code/session_01UMk9eURri2AqBbsMNwLQVJ


src/tabpfn_client/api_models.py
```diff
@@ -216,6 +216,9 @@ class ThinkingConfig(BaseModel):
     effort: Annotated[ThinkingEffort | str, Field(union_mode="left_to_right")] | None = None
     timeout_secs: float | None = None
     metric: str | None = None
+    group_col: str | list[str] | None = None
```
brendan-priorlabs, line 219: @ggprior, this does match the interface that @Innixma originally proposed here https://prior-labs.slack.com/archives/C08R6A4N54H/p1788450717918369?thread_ts=1787590288.540279&cid=C08R6A4N54H. Some reason we're okay with only supporting strings right now?

brendan-priorlabs (dismissed):
Stamped, but please see comment

src/tabpfn_client/api_models.py
ggprior: Right now I'm requiring X to be a pandas dataframe with named columns, validating this below in _validate_group_columns. An integer (column index) could also be acceptable here but makes the server plumbing a little more complicated because we need to keep track of the columns in the preprocessing stack. Checking how feasible this is

brendan-priorlabs: Sounds good. My read is that this isn't critical as long as we didn't change the API to require Pandas for everything, just this new feature. We can widen the interface later

brendan-priorlabs (approved):

PR #396, opened 2026-09-22
Validate upload caps and defer subsampled contexts to the server

adrian-prior:
Generated by Codex

Validate training uploads against the upload cap returned by the server so valid subsampled uploads are not rejected against the per-estimator cap. Keep static shape and output checks in the client, and defer subsampled context validation and context-dependent prediction limits to the server. This avoids duplicating TabPFN's subsampling rules in the client; context errors can be reported after upload.

Validation: 20 validation, estimator, and group-column tests passed. The full configured Trunk gate passed across 57 files.


src/tabpfn_client/estimator.py
```diff
@@ -962,9 +964,46 @@ def _validate_targets(self, y) -> None:
             raise ValueError("Input y contains NaN.")
 
 
+def effective_context_rows(n_rows: int, config: TabPFNConfig | None) -> int:
```
ggprior: Oh boy, this proposes to duplicate the logic essentially 3 times, once here, once in tabpfn-server, once in TabPFN. All need to be kept in sync.
Is there a case for at least avoiding it here (and say have the server fail fast)?

ggprior (dismissed):
Let's check if we can avoid the duplication. tabpfn-common-utils or server-side validation?
Otherwise LGTM pending this double check.

src/tabpfn_client/estimator.py
adrian-prior: Remove it here. It seems like the server would fail quickly. I will look into deduplicating the rest, too.

src/tabpfn_client/api_models.py
```diff
@@ -186,6 +186,7 @@ class FitTimings(BaseModel):
 class ModelLimit(BaseModel):
     train_set_max_rows: int
     train_set_max_cells: int
+    train_set_max_upload_cells: int | None = None
```
simo-prior: This is slightly different from what we have on the server, you can copy-paste the whole class from here: https://github.com/PriorLabs/tabpfn-server/blob/main/libs/prior/sdks/gapi/models_gen.py#L213

src/tabpfn_client/estimator.py
```diff
@@ -963,8 +965,10 @@ def _validate_targets(self, y) -> None:
 
 
 def validate_train_set(
-    X: pd.DataFrame | np.ndarray, y: pd.Series | np.ndarray | None = None
-):
+    X: pd.DataFrame | np.ndarray,
+    y: pd.Series | np.ndarray | None = None,
+    tabpfn_config: TabPFNConfig | None = None,
```
simo-prior: I would only pass `model_path` here.

src/tabpfn_client/estimator.py
```diff
@@ -1018,6 +1027,7 @@ def validate_test_set(
     model_path: str | None = None,
     train_rows: int | None = None,
     split_full_output: bool = False,
+    tabpfn_config: TabPFNConfig | None = None,
```
simo-prior: Same criterion as above, I'd only pass `inference_config`

src/tabpfn_client/estimator.py
```diff
@@ -989,9 +997,10 @@ def validate_train_set(
             f"The number of train columns ({X.shape[1]}) exceeds the maximum of {limit.max_cols}."
         )
     n_cells = X.shape[0] * X.shape[1]
-    if n_cells > limit.train_set_max_cells:
+    upload_limit = limit.train_set_max_upload_cells or limit.train_set_max_cells
```
simo-prior: With the right class definition this will never be None (else limit.train_set_max_cells will also be).

src/tabpfn_client/estimator.py
```diff
@@ -976,9 +980,13 @@ def validate_train_set(
     if api_settings is None:
         return
 
-    # We don't yet know which model will be used, so we use the most permissive limit
-    # across all models.
-    limit = api_settings.max_model_limit
+    limit = (
+        _limit_for_model_path(tabpfn_config.model_path)
```
simo-prior: `limit = _limit_for_model_path(tabpfn_config.model_path)` should be sufficient here, if model_path is None it automatically resolves to the max_model_limit. Still need to return if the returned limit is None ofc.

simo-prior (commented):
.

simo-prior (approved):
Thanks!

